PR #41, opened 2016-07-15
Implemetation

ebachelet:
Hi guys

I have implemented herder_checker qc and test accordingly. Will I pass the severe Curtis criteria?

### DIFF
--- a/banzai/qc/header_checker.py
+++ b/banzai/qc/header_checker.py
@@ -1,45 +1,54 @@
+"""
+This module performs controls on several critical parameters of image header, contains in
+header_expected_format dictionnary.
+@author:ebachelet
+"""
 from banzai.stages import Stage
 from banzai import logs
 
-import os
-
-
 class HeaderSanity(Stage):
     """
-    Check the header of the image.
-
+       This class contains the needed function/definitions to performs the header check.
     """
 
     def __init__(self, pipeline_context):
         super(HeaderSanity, self).__init__(pipeline_context)
 
-        self.Header_expected_format = {'RA': str, 'DEC': str, 'CAT-RA': str, 'CAT-DEC': str, 'OFST-RA': str,
-                                       'OFST-DEC': str, 'TPT-RA': str, 'TPT-DEC': str, 'PM-RA': str, 'PM-DEC': str,
-                                       'CRVAL1': float, 'CRVAL2': float, 'CRPIX1': int, 'CRPIX2': int,
-                                       'EXPTIME': float}
+        self.header_expected_format = {'RA': str, 'DEC': str, 'CAT-RA': str, 'CAT-DEC': str,
+                                       'OFST-RA': str, 'OFST-DEC': str, 'TPT-RA': str,
+                                       'TPT-DEC': str, 'PM-RA': str, 'PM-DEC': str,
+                                       'CRVAL1': float, 'CRVAL2': float, 'CRPIX1': int,
+                                       'CRPIX2': int, 'EXPTIME': float}
 
     @property
     def group_by_keywords(self):
         return None
 
     def do_stage(self, images):
-
+        """ Performs several checks on the header, see the relative functions.
+        :param list images: a list of image object.
+        :returns: the list of image object after check
+        :rtype: list
+        """
         for image in images:
 
 
-            for keyword in self.Header_expected_format.keys():
+            for keyword in self.header_expected_format.keys():
                 self.check_header_keyword_present(keyword, image)
                 self.check_header_format(keyword, image)
-                self.check_header_NA(keyword, image)
+                self.check_header_na(keyword, image)
 
-            self.check_RA_range(image)
-            self.check_DEC_range(image)
-            self.check_EXPTIME_value(image)
+            self.check_ra_range(image)
+            self.check_dec_range(image)
+            self.check_exptime_value(image)
 
         return images
 
     def check_header_keyword_present(self, keyword, image):
-
+        """ Return a warning if the keyword is not in the image header.
+        :param str keyword: the stry keyword you look inside the header
+        :param object image: an image object
+        """
         header = image.header
 
         logging_tags = logs.image_config_to_tags(image, self.group_by_keywords)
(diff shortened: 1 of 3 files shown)


banzai/tests/test_stats.py
```diff
@@ -529,15 +529,22 @@ def test_mad_3d_axis_1_mask_none():
...
         a = np.random.normal(mean, sigma, size=(size1, size2, size3))
+        b = a.copy()
+        
         expected = np.median(np.abs(a.astype(np.float32) - np.median(a.astype(np.float32), axis=2).reshape(size1, size2, 1)), axis=2)
-        actual = stats.median_absolute_deviation(a, axis=2)
-        np.testing.assert_allclose(actual, expected.astype(np.float32), atol=1e-4)
+        actual = stats.median_absolute_deviation(b, axis=2)
+        print(a[:,:,0])
+        print(np.median(a, axis=2)[0,0])
+        print(actual[0,0], expected[0,0])
```
cmccully: Tests should not print anything.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def do_stage(self, images):
+        """ Performs several checks on the header, see the relative functions.
+        :param list images: a list of image object.
+        :returns: the list of image object after check
+        :rtype: list
+        """
+        for image in images:
+
+
```
cmccully: Only skip one line between these blocks.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+            self.check_ra_range(image)
+            self.check_dec_range(image)
+            self.check_exptime_value(image)
+
+        return images
+
+    def check_header_keyword_present(self, keyword, image):
+        """ Return a warning if the keyword is not in the image header.
+        :param str keyword: the stry keyword you look inside the header
```
cmccully: This format is for the docstring is fine, but I tend to try to follow the astropy convention which is slightly different.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
+"""
+This module performs controls on several critical parameters of image header, contains in
```
cmccully: I would rephrase this to be something like "This module performs basic sanity checks that the main image header keywords are the correct format and validates their values."

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+header_expected_format dictionnary.
+@author:ebachelet
+"""
+from banzai.stages import Stage
+from banzai import logs
+
+class HeaderSanity(Stage):
+    """
+       This class contains the needed function/definitions to performs the header check.
```
cmccully: This docstring does not really have any information in it. Maybe something like "Stage to validate important header keywords" or something like that.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+
+    def __init__(self, pipeline_context):
+        super(HeaderSanity, self).__init__(pipeline_context)
+
+        self.header_expected_format = {'RA': str, 'DEC': str, 'CAT-RA': str, 'CAT-DEC': str,
+                                       'OFST-RA': str, 'OFST-DEC': str, 'TPT-RA': str,
+                                       'TPT-DEC': str, 'PM-RA': str, 'PM-DEC': str,
```
cmccully: I am nervous to include all of the variations of RA and Dec here because they may change over time. I don't really have a good solution to block against that though.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+                                       'CRVAL1': float, 'CRVAL2': float, 'CRPIX1': int,
+                                       'CRPIX2': int, 'EXPTIME': float}
+
+    @property
+    def group_by_keywords(self):
+        return None
+
+    def do_stage(self, images):
+        """ Performs several checks on the header, see the relative functions.
```
cmccully: Maybe: "Run stage to validate header"

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+        :param list images: a list of image object.
```
cmccully: list of banzais.images.Image

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+        :returns: the list of image object after check
```
cmccully: list of banzai.images.Image
List of validated images.

Do we ever reject images? If so we should mention that here.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+                self.check_header_na(keyword, image)
+
+            self.check_ra_range(image)
+            self.check_dec_range(image)
+            self.check_exptime_value(image)
+
+        return images
+
+    def check_header_keyword_present(self, keyword, image):
+        """ Return a warning if the keyword is not in the image header.
```
cmccully: Not return. Logs.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+        :param str keyword: the stry keyword you look inside the header
+        :param object image: an image object
+        """
+        header = image.header
+
+        logging_tags = logs.image_config_to_tags(image, self.group_by_keywords)
+
+        if keyword not in header:
+            sentence = 'WARNING : The header key ' + keyword + ' is not in image header!'
```
cmccully: Don't have warning in the message itself. Use logger.warn and that will produce the warning message correctly.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+
+            self.logger.error(sentence, extra=logging_tags)
```
cmccully: Use warning if you want it to be a warning. Or error if you want it to be an error.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+
+            return 'WARNING'
```
cmccully: You do not need to return anything here.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+        logging_tags = logs.image_config_to_tags(image, self.group_by_keywords)
+
+        if not isinstance(header_value, self.header_expected_format[keyword]):
+            sentence = ('The header key ' + keyword + ' got an unexpected format : ' + type(
+                header_value).__name__ + ' in place of ' + self.header_expected_format[
+                    keyword].__name__)
+
+            self.logger.error(sentence, extra=logging_tags)
+
+        return
```
cmccully: You do not need a return if you return None.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+        if keyword not in header:
+            sentence = 'WARNING : The header key ' + keyword + ' is not in image header!'
+
+            self.logger.error(sentence, extra=logging_tags)
+
+            return 'WARNING'
+        return
+
+    def check_header_format(self, keyword, image):
+        """ Return a warning if the keyword is not the expected type
```
cmccully: Again log a warning.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+            in the image header.
+        :param str keyword: the stry keyword you look inside the header
```
cmccully: stry?

cmccully: Also, maybe just "header keyword to check"

banzai/qc/header_checker.py
cmccully: Use this throughout.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+                header_value).__name__ + ' in place of ' + self.header_expected_format[
+                    keyword].__name__)
+
+            self.logger.error(sentence, extra=logging_tags)
+
+        return
+
+    def check_header_na(self, keyword, image):
+        """ Return a warning if the keyword is 'N/A' instead of the
+            expected type in the image header.
```
cmccully: This is good, but all bias frames and dark frames have RA and Dec as N/A. So we don't want to produce warning for every single one of those calibration files.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+        :param str keyword: the stry keyword you look inside the header
```
cmccully: See comment above.

banzai/qc/header_checker.py
```diff
@@ -0,0 +1,155 @@
...
+        if isinstance(header_value, str):
+
+            if 'N/A' in header_value:
+                sentence = 'The header key ' + keyword + ' got the unexpected value : N/A'
+                self.logger.error(sentence, extra=logging_tags)
+
+        return
+
+    def check_ra_range(self, image):
+        """ Return an error if the keyword right_ascension is not inside
```
cmccully: Again this doesn't return an error. See below as well. This seems like some copy and pasted code. Could you abstract this a bit into a single function?

banzai/tests/test_header_checker.py
```diff
@@ -0,0 +1,127 @@
...
+    ny = 103
+
+    images = [FakeImage(nx=nx, ny=ny) for x in range(6)]
+
+    for image in images:
+
+        tester.check_header_keyword_present('EXPTIME', image)
+        assert tester.logger.error.called
+
+    assert len(images) == 6
```
cmccully: This would be where we could check if images are being rejected correctly.

banzai/tests/test_header_checker.py
```diff
@@ -0,0 +1,127 @@
...
+    tester.logger.error = mock.MagicMock()
+    nx = 101
+    ny = 103
+
+    images = [FakeImage(nx=nx, ny=ny) for x in range(6)]
+    for image in images:
+        image.header['RA'] = 20
+
+        tester.check_header_format('RA', image)
+        assert tester.logger.error.called
```
cmccully: This makes me nervous a bit. What if we decide to switch RA and Dec to decimal instead of segsidecimal?

banzai/tests/test_stats.py
ebachelet: OK done, this comes from an old discussion we had on low number statistics

banzai/qc/header_checker.py
ebachelet: I think we decided a while ago to not reject anything from this module, because some info could be found again easily, examples date of observation or RA DEC

banzai/qc/header_checker.py
ebachelet: Does these BIAS and DARK have specificity inside OBSTYPE for example? If yes, I can put an exception.

banzai/tests/test_header_checker.py
ebachelet: See my first comment.

banzai/tests/test_header_checker.py
ebachelet: We decided to not reject frame form this module. Just to indicates logs. We can change this if needed now, but this needs discussions.

banzai/qc/header_checker.py
```diff
@@ -58,15 +71,19 @@ def check_header_keyword_present(self, keyword, image):
 
             self.logger.error(sentence, extra=logging_tags)
```
cmccully: Is this supposed to to log an error or warning?

PR #77, opened 2017-02-13
Various fixes, increased test coverage, and added multi-extension fits support for raw files.

cmccully:


banzai/images.py
```diff
@@ -35,7 +38,11 @@ def __init__(self, pipeline_context, filename=None, data=None, header={}, bpm=No
         self.nx = header.get('NAXIS1')
         self.ny = header.get('NAXIS2')
 
-        self.gain = header.get('GAIN')
+        if len(self.extension_headers) > 0 and 'GAIN' in self.extension_headers[0]:
```
ire-and-curses: Could write the len() check more simply as just

if self.extension_headers

banzai/main.py
```diff
@@ -26,6 +26,49 @@
...
+                  trim.Trimmer,
+                  bias.BiasSubtractor,
+                  dark.DarkSubtractor,
+                  flats.FlatDivider,
+                  photometry.SourceDetector,
+                  astrometry.WCSSolver,
+                  pointing.PointingTest]
+
+
+def get_stages_todo(last_stage, extra_stages=[]):
```
ire-and-curses: This isn't going to do what you want. The [] is evaluated at compile time, not on each function invocation. This is called the "mutable default arguments" problem. Best practice looks clunky, but recommended is

def get_stages_todo(last_stage, extra_stages=None):
    if not extra_stages:
        extra_stages = []

ire-and-curses: See e.g. https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

banzai/mosaic.py
```diff
@@ -26,37 +26,49 @@ def do_stage(self, images):
...
-                image.header['NAXIS'] = 2
-                image.header.pop('NAXIS3')
-
+                update_naxis_keywords(image, nx, ny)
                 self.logger.info('Mosaiced image', extra=logging_tags)
         return images
 
 
+def update_naxis_keywords(image, nx, ny):
+    if 'NAXIS3' in image.header.keys():
```
ire-and-curses: in defaults to looking at keys, so the "Pythonic" way to write this is just

if 'NAXIS3' in image.header:

banzai/munge.py
```diff
@@ -18,83 +21,12 @@ def do_stage(self, images):
         for image in images:
             telescope = dbs.get_telescope(image.telescope_id,
                                           db_address=self.pipeline_context.db_address)
-            # TODO: Currently we only support 1x1 Sinistro frames and only support 1x1 frames.
-            # TODO: 2x2 frames cannot use hard coded values because we read out more pixels.
+
             if 'sinistro' in telescope.camera_type.lower():
```
ire-and-curses: The whole of the following block is pretty nasty. The nastiness comes from the intimate coupling of the data values and the structure of the logic. I would look at a possible rewrite splitting the data into one or two dicts, and then indexing into it as necessary within the logic (I can explain more in person if helpful).

banzai/munge.py
```diff
@@ -110,25 +42,134 @@ def do_stage(self, images):
...
+    """
+    # TODO Add support for other binnings
+    supported = True
+
+    tags = logs.image_config_to_tags(image, None)
+    logs.add_tag(tags, 'filename', image.filename)
+    if image.header['CCDSUM'] != '1 1':
+        supported = False
+        logger.error('Non-supported Sinistro mode', logging_tags=tags)
+    if image.instrument not in crosstalk_coefficients.keys():
```
ire-and-curses: keys

PR #84, opened 2017-03-03
Add defaults to context object.

cmccully:


cmccully:
Will create a new PR that will address this issue.

PR #85, opened 2017-03-06
Celery

cmccully:
Refactoring the code to use celery to manage the parallel workers. This should solve the heartbeat issue from rabbitmq having a load balancer in front of it.


banzai/tasks.py
```diff
@@ -0,0 +1,71 @@
...
+                                max_tries=pipeline_context.max_preview_tries):
+        stages_to_do = stages.get_stages_todo()
+
+        logging_tags = {'tags': {'filename': os.path.basename(path)}}
+        logger.info('Running preview reduction on {}'.format(path), extra=logging_tags)
+        pipeline_context.filename = os.path.basename(path)
+        pipeline_context.raw_path = os.path.dirname(path)
+
+        # Increment the number of tries for this file
+        dbs.increment_preview_try_number(path, db_address=pipeline_context.db_address)
```
Fingel: Just FYI you have access to self.request.retries (int number of times this task as retried) and self.retry() which re enqueues the task (takes a few helpful arguments as well)

cmccully: Cool thanks!

cmccully:
Will be closed in #189

PR #96, opened 2018-01-02
remove fits2fits option as it is deprecated in astrometry.net at leas…

drhaz:
…t as of version 0.72


drhaz:
Bug fix in response to change of command line option in astrometry.net. See
https://github.com/dstndstn/astrometry.net/issues/100

PR #106, opened 2018-05-08
Standardize elasticsearch field names

turnerm:


banzai/qc/pattern_noise.py
```diff
@@ -39,7 +38,9 @@ def do_stage(self, images):
                 images_to_remove.append(image)
             else:
                 self.logger.info('No pattern noise found.', extra=logging_tags)
-            self.save_qc_results({'PatternNoise': pattern_noise_is_bad}, image)
+            self.save_qc_results({'pattern_noise.failed': pattern_noise_is_bad,
+                                  'pattern_noise.snr_threshold': self.SNR_THRESHOLD,
+                                  'pattern_noise.pixel_threshold': self.PIXEL_THRESHOLD}, image)
```
cmccully: This isn't really a number of pixels in the normal sense. This is actually the number of bins in a stacked 2d FFT. Maybe we should call this freq_bins_threshold?

banzai/qc/pointing.py
```diff
@@ -61,9 +61,11 @@ def do_stage(self, images):
...
                 self.logger.warning('Pointing offset exceeds threshhold',
                                     extra=self.logging_tags)
-            self.save_qc_results({'PointingSevere': pointing_severe,
-                                  'PointingWarning': pointing_warning,
-                                  'pointing_offset': angular_separation},
+            self.save_qc_results({'pointing.severe.failed': pointing_severe,
+                                  'pointing.severe.threshold': self.SEVERE_THRESHOLD,
+                                  'pointing.warning.failed': pointing_warning,
+                                  'pointing.warning.threshold': self.WARNING_THRESHOLD,
+                                  'pointing.offset': angular_separation},
```
cmccully: I don't really like failed here. Maybe triggered? We should think a bit about what happens if we have tiered thresholds like here with a warning then a severe.

banzai/qc/saturation.py
```diff
@@ -35,16 +35,18 @@ def do_stage(self, images):
             logs.add_tag(logging_tags, 'SATFRAC', saturation_fraction)
             logs.add_tag(logging_tags, 'threshold', self.SATURATION_THRESHOLD)
             self.logger.info('Measured saturation fraction.', extra=logging_tags)
-            saturated = saturation_fraction >= self.SATURATION_THRESHOLD
-            if saturated:
+            is_saturated = saturation_fraction >= self.SATURATION_THRESHOLD
+            qc_results = {'saturated.failed': is_saturated,
```
cmccully: As a general question, do we want to have a failed=false for every check or just only have a failed tag if it actually failed?

banzai/qc/sinistro_1000s.py
```diff
@@ -36,13 +36,16 @@ def do_stage(self, images):
             logs.add_tag(logging_tags, 'FRAC1000', fraction_1000s)
             logs.add_tag(logging_tags, 'threshold', self.THOUSANDS_THRESHOLD)
             has_1000s_error = fraction_1000s > self.THOUSANDS_THRESHOLD
+            qc_results = {'sinistro_thousands.failed': has_1000s_error,
+                          'sinistro_thousands.fraction': fraction_1000s,
+                          'sinistro_thousands.threshold': self.THOUSANDS_THRESHOLD}
             if has_1000s_error:
                 self.logger.error('Image is mostly 1000s. Rejecting image', extra=logging_tags)
+                qc_results['rejected'] = True
```
cmccully: This ties into my question above about whether the failed flags should be there or not. Should we be consistent?

banzai/tests/test_saving_qc.py
```diff
@@ -11,6 +11,7 @@ def test_format_qc_results_basic_info():
     assert results['instrument'] == image.instrument
     assert results['dayobs'] == image.epoch
     assert results['@timestamp'] == image.dateobs
+    assert results['obstype'].upper() == image.obstype
```
cmccully: Should obstype be upper case since it came from the FITS header? Or should we use full words here?

banzai/qc/header_checker.py
```diff
@@ -213,6 +216,7 @@ def check_exptime_value(self, image):
             if exptime_zero:
                 sentence = 'The header EXPTIME key got the unexpected value : 0.0'
                 self.logger.error(sentence, extra=logging_tags)
-            self.save_qc_results({"HeaderExptimeZero": exptime_zero}, image)
+            qc_results["header.exptime.zero.failed"] = exptime_zero
         except:
             pass
+        self.save_qc_results(qc_results, image)
```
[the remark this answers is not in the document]
turnerm: Done

banzai/qc/pattern_noise.py
turnerm: I agree, I've changed it to your suggested name

banzai/qc/pointing.py
turnerm: I agree, I think "triggered" works better for the warnings. Alternatively, I could take out the distinctions between warning and severe explicitly and just use "pointing.warning" for warning and "pointing.failed" for severe.

banzai/qc/saturation.py
turnerm: The former (failed=false for every check). I would argue to keep it this way, since it indicates whether or not a check was performed at all.

banzai/qc/sinistro_1000s.py
turnerm: We could have a default "rejected=false" flag, however in contrast to the "failed=false" flag I don't think it adds much information, and we would have to make sure to plumb it through properly (perhaps as an image attribute?)

banzai/tests/test_saving_qc.py
turnerm: I put obstype into elasticsearch as lowercase to try and stick with elasticsearch naming conventions, but actually since this is the field value and not the field name then it's probably more correct to just keep whatever comes out of the header as-is. I've removed the case changes.

banzai/qc/pointing.py
cmccully: I like your second suggestion of using point.warning and pointing.failed.

banzai/qc/saturation.py
cmccully: Ok.

banzai/qc/sinistro_1000s.py
cmccully: Ok. I think this is fine how it is.

banzai/qc/pointing.py
turnerm: Done

PR #107, opened 2018-05-08
Fixed bugs and problems in header checker 

turnerm:


banzai/qc/header_checker.py
```diff
@@ -20,11 +20,11 @@ class HeaderSanity(Stage):
     def __init__(self, pipeline_context):
         super(HeaderSanity, self).__init__(pipeline_context)
 
-        self.header_expected_format = {'RA': str, 'DEC': str, 'CAT-RA': str, 'CAT-DEC': str,
-                                       'OFST-RA': str, 'OFST-DEC': str, 'TPT-RA': str,
-                                       'TPT-DEC': str, 'PM-RA': str, 'PM-DEC': str,
-                                       'CRVAL1': float, 'CRVAL2': float, 'CRPIX1': int,
-                                       'CRPIX2': int, 'EXPTIME': float}
+        self.header_keywords_list = ['RA', 'DEC', 'CAT-RA', 'CAT-DEC',
```
cmccully: maybe expected_header_keyword list?

banzai/qc/header_checker.py
```diff
@@ -45,144 +45,99 @@ def do_stage(self, images):
...
+            self.check_ra_range(image, bad_keywords)
+            self.check_dec_range(image, bad_keywords)
+            self.check_exptime_value(image, bad_keywords)
         return images
 
-    def check_header_keyword_present(self, keywords, image):
-        """ Logs an error if the keyword is not in the image header.
+    def check_keywords_missing_or_na(self, image):
+        """ Logs an error if the keyword right_ascension is not inside
+            the expected range (0<ra<360 degrees) in the image header.
```
cmccully: Is this supposed to be here in the comment? ra seems like it would be somewhere else.

banzai/qc/header_checker.py
```diff
@@ -45,144 +45,99 @@ def do_stage(self, images):
...
         """
-
         logging_tags = logs.image_config_to_tags(image, self.group_by_keywords)
+        qc_results = {}
         missing_keywords = []
-        for keyword in keywords:
+        na_keywords = []
+        for keyword in self.header_keywords_list:
             if keyword not in image.header:
                 sentence = 'The header key ' + keyword + ' is not in image header!'
```
cmccully: I like the .format method here instead of +

banzai/qc/header_checker.py
```diff
@@ -45,144 +45,99 @@ def do_stage(self, images):
...
+                self.logger.error(sentence, extra=logging_tags)
                 missing_keywords.append(keyword)
+            elif image.header[keyword] == 'N/A':
```
cmccully: Won't bias frames always get an RA of N/A now that we are running this on calibration frames. We should probably have a rule here about obstypes.

banzai/qc/header_checker.py
```diff
@@ -45,144 +45,99 @@ def do_stage(self, images):
...
+                sentence = 'The header key ' + keyword + ' got the unexpected value : N/A'
                 self.logger.error(sentence, extra=logging_tags)
-        any_keywords_missing = True if len(missing_keywords) else False
-        qc_results = {"header.keywords.missing.failed": any_keywords_missing}
-        if any_keywords_missing:
+                na_keywords.append(keyword)
+        are_keywords_missing = True if len(missing_keywords) else False
```
cmccully: why not keywords_are_missing = len(missing_keywords) > 0

banzai/qc/header_checker.py
```diff
@@ -45,144 +45,99 @@ def do_stage(self, images):
...
+        are_keywords_na = True if len(na_keywords) else False
```
cmccully: See above.

banzai/qc/header_checker.py
```diff
@@ -45,144 +45,99 @@ def do_stage(self, images):
...
+        if 'CRVAL1' not in bad_keywords:
+            logging_tags = logs.image_config_to_tags(image, self.group_by_keywords)
+            ra_value = image.header['CRVAL1']
+            is_bad_ra_value = (ra_value > self.RA_MAX) | (ra_value < self.RA_MIN)
+            if is_bad_ra_value:
                 sentence = 'The header CRVAL1 key got the unexpected value : ' + str(ra_value)
                 self.logger.error(sentence, extra=logging_tags)
-            self.save_qc_results({"header.ra.failed": bad_ra_value,
+            self.save_qc_results({"header.ra.failed": is_bad_ra_value,
                                   "header.ra.value": ra_value}, image)
```
cmccully: Should you save the range used? Or is 0 to 360 obvious?

banzai/qc/header_checker.py
```diff
@@ -45,144 +45,99 @@ def do_stage(self, images):
...
+        if 'CRVAL2' not in bad_keywords:
+            logging_tags = logs.image_config_to_tags(image, self.group_by_keywords)
+            dec_value = image.header['CRVAL2']
+            is_bad_dec_value = (dec_value > self.DEC_MAX) | (dec_value < self.DEC_MIN)
+            if is_bad_dec_value:
                 sentence = 'The header CRVAL2 key got the unexpected value : ' + str(dec_value)
                 self.logger.error(sentence, extra=logging_tags)
-            self.save_qc_results({"header.dec.failed": bad_dec_value,
+            self.save_qc_results({"header.dec.failed": is_bad_dec_value,
                                   "header.dec.value": dec_value}, image)
```
cmccully: Same comment about range?

cmccully:
banzai/tests/test_header_checker.py:114:42: W292 no newline at end of file

banzai/qc/header_checker.py
turnerm: Done

banzai/qc/header_checker.py
turnerm: Fixed

banzai/qc/header_checker.py
turnerm: Updated all strings to use format method

banzai/qc/header_checker.py
turnerm: Bias frames do have numerical `RA` and `DEC` values, although their `CAT-RA` and `CAT-DEC` values are usually set to `NaN`. But it is `N/A` values that specifically indicate a problem (see e.g. issue [9784](https://issues.lco.global/issues/9784))

banzai/qc/header_checker.py
turnerm: Fixed

banzai/qc/header_checker.py
turnerm: Fixed

banzai/qc/header_checker.py
turnerm: Partly because it's obvious, but also partly because it's set by the definition of RA and will never change, so there is no need to keep track of it.

banzai/qc/header_checker.py
```diff
@@ -45,178 +45,120 @@ def do_stage(self, images):
...
+            bad_keywords = self.check_keywords_missing_or_na(image)
+            self.check_ra_range(image, bad_keywords)
+            self.check_dec_range(image, bad_keywords)
+            self.check_exptime_value(image, bad_keywords)
         return images
 
-    def check_header_keyword_present(self, keywords, image):
-        """ Logs an error if the keyword is not in the image header.
+    def check_keywords_missing_or_na(self, image):
+        """ Logs an error if the keyword is missing or 'N/A'.
```
cmccully: Can you add a comment about the discussion about N/A vs Nan here.

PR #108, opened 2018-05-10
Hotfix/pattern noise

turnerm:
Images with detected pattern noise are no longer removed, and calibration comparison is no longer run for master calibration creation


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+0.9.3 (2018-05-10)
+------------------
+- Hotfix
```
cmccully: I would note that both of these are temporary states.

PR #109, opened 2018-05-15
Fix/calibration checker

turnerm:
Some small bug fixes and updates to the calibration checker: 

- Since the default behaviour in `stages.CalibrationComparer` is to divide `bad_pixel_fraction` by `noise`, I changed the default noise model to return ones rather than zeros to avoid zero division. 
- Changed how a missing master calibration file is handled for the checker (previously threw an error)
- Updated `FLAT` to `SKYFLAT` in `stages.py` (the comparison has not been running so far)

### DIFF
--- a/banzai/stages.py
+++ b/banzai/stages.py
@@ -226,4 +226,4 @@ class CalibrationComparer(ApplyCalibration):
 
     @abc.abstractmethod
     def noise_model(self, image):
-        return np.zeros(image.data.size)
+        return np.ones(image.data.size)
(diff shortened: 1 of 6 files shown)


banzai/stages.py
```diff
@@ -160,10 +160,11 @@ def do_stage(self, images):
 
             if master_calibration_filename is None:
                 self.on_missing_master_calibration(logging_tags)
-
-            master_calibration_image = Image(self.pipeline_context,
+            else:
```
cmccully: I think this will break things down the road. If there is no master calibration it makes a default image object that can be used later down the line. This would cause a bunch of extra exceptions.

turnerm: Reverted

PR #110, opened 2018-05-17
Hotfix to fix retrying making preview images

cmccully:


banzai/dbs.py
```diff
@@ -365,9 +365,18 @@ def need_to_make_preview(path, db_address=_DEFAULT_DB, max_tries=5):
     need_to_process = False
```
turnerm: (line 364) I think this comment needs updating

cmccully: Done.

PR #116, opened 2018-07-03
Feature/refactor image3d

turnerm:


banzai/images.py
```diff
@@ -131,6 +131,40 @@ def write_catalog(self, filename, nsources=None):
...
+        return data_is_3d
+
+    def get_n_amps(self):
+        if self.data_is_3d():
+            n_amps = self.data.shape[0]
+        else:
+            n_amps = 1
+        return n_amps
+
+    def get_inner_image_section(self, edge_fraction_denominator=4):
```
cmccully: edge_fraction_denominator is a bit of a confusing name. How about something like section_denominator?

banzai/images.py
```diff
@@ -131,6 +131,40 @@ def write_catalog(self, filename, nsources=None):
...
+        if self.data_is_3d():
+            logger.error("Cannot get inner section of a 3D image",
+                         extra={'tags': {'filename': self.filename}})
+            inner_section = self.data
+        else:
+            # Note that (-a//b) != -(a//b) if a is odd
+            try:
+                inner_nx = self.nx // edge_fraction_denominator
+                inner_ny = self.ny // edge_fraction_denominator
+                inner_section = self.data[inner_ny: -inner_ny, inner_nx: -inner_nx]
```
cmccully: I'm confused why this try is necessary?

banzai/tests/test_image.py
```diff
@@ -1,7 +1,51 @@
 from banzai.images import Image
-from banzai.tests.utils import FakeContext
+from banzai.tests.utils import FakeContext, FakeImage
+import numpy as np
 
+rs = np.random.RandomState(0)
```
cmccully: This has no real affect in pytest. This should be a fixture like in test_dark_normalizer.

banzai/tests/test_image.py
```diff
@@ -1,7 +1,51 @@
...
+    test_image = FakeImage(n_amps=n_amps)
+    assert test_image.get_n_amps() == n_amps
+
+
+def test_get_inner_quarter_default():
+    test_image = FakeImage()
+    test_image.data = rs.randint(0, 1000, size=test_image.data.shape)
+    # get inner quarter manually
+    inner_nx = test_image.nx//4
+    inner_ny = test_image.ny//4
```
cmccully: PEP8.

banzai/images.py
turnerm: I changed it to multiply `nx` and `ny` and renamed it `fractional_edge_width`.

banzai/images.py
turnerm: In case `nx`, `ny` or `data` are not set, then it will just return `data` as is and log an error, rather than crashing.

banzai/tests/test_image.py
turnerm: Fixed

banzai/tests/test_image.py
turnerm: Fixed

PR #117, opened 2018-07-18
read_images would only act properly on images without a BPM.

gmbrandt:
Fixed banzai.images.read_images so that if image.bpm is not None, then read_images will fill out the image list as it should. Prior, it would return an empty list which would cause main.run to act on an empty list of images and do nothing.


cmccully:
Fixed via #120

PR #118, opened 2018-07-18
Fix/pattern noise

turnerm:


banzai/main.py
```diff
@@ -85,7 +85,7 @@ def get_preview_stages_todo(image_suffix):
                                  extra_stages=[dark.DarkNormalizer, dark.DarkComparer])
     elif image_suffix == 'f00.fits':
         stages = get_stages_todo(last_stage=dark.DarkSubtractor,
-                                 extra_stages=[flats.FlatNormalizer, flats.FlatComparer])
+                                 extra_stages=[flats.FlatNormalizer, qc.PatternNoiseDetector, flats.FlatComparer])
```
cmccully: Why do we want to run this so late in the process? Why not run it earlier?

banzai/main.py
```diff
@@ -138,7 +138,9 @@ def make_master_dark_console():
 
 
 def make_master_flat(pipeline_context):
-    stages_to_do = get_stages_todo(dark.DarkSubtractor, extra_stages=[flats.FlatNormalizer, flats.FlatMaker])
+    stages_to_do = get_stages_todo(dark.DarkSubtractor, extra_stages=[flats.FlatNormalizer,
+                                                                      qc.PatternNoiseDetector,
+                                                                      flats.FlatMaker, ])
```
cmccully: See above.

banzai/qc/pattern_noise.py
```diff
@@ -1,16 +1,17 @@
...
 
 
 class PatternNoiseDetector(Stage):
     # Signal to Noise threshold to raise an alert
-    SNR_THRESHOLD = 15.0
-    # Number of pixels that need to be above the S/N threshold to raise an alert
-    FREQ_BINS_THRESHOLD = 5
+    SNR_THRESHOLD = 10.0
+    # The maximum allowed standard deviation of the peak centres
+    STD_THRESHOLD = 1.0
```
cmccully: Is there a better name for this?

banzai/qc/pattern_noise.py
```diff
@@ -21,49 +22,197 @@ def group_by_keywords(self):
...
             logs.add_tag(logging_tags, 'filename', os.path.basename(image.filename))
             logs.add_tag(logging_tags, 'snr_threshold', self.SNR_THRESHOLD)
-            logs.add_tag(logging_tags, 'freq_bins_threshold', self.FREQ_BINS_THRESHOLD)
+            logs.add_tag(logging_tags, 'std_threshold', self.STD_THRESHOLD)
+
+            # If the data is a cube, then run on each extension individually
             if image.data_is_3d():
-                pattern_noise_is_bad = any([check_for_pattern_noise(data, self.SNR_THRESHOLD, self.FREQ_BINS_THRESHOLD)
-                                            for data in image.data])
+                pattern_noise_is_bad = any([self.check_for_pattern_noise(data) for data in image.data])
```
cmccully: If this runs after the stitching, will it ever be a cube?

banzai/qc/pattern_noise.py
```diff
@@ -21,49 +22,197 @@ def group_by_keywords(self):
...
+        """
+
+        trimmed_data = trim_image_edges(data)
+        power_2d = get_2d_power_band(trimmed_data)
+        snr = compute_snr(power_2d)
+        convolved_snr = convolve_snr_with_wavelet(snr)
+        peak_maxima, std_maxima = get_peak_parameters(convolved_snr)
+
+        # Check that all peaks are above threshold and that the peak center standard deviation is small.
+        has_pattern_noise = (peak_maxima > self.SNR_THRESHOLD).all() and std_maxima < self.STD_THRESHOLD
```
cmccully: What happens if no peaks are found?

banzai/tests/test_pattern_noise_qc.py
```diff
@@ -83,3 +86,56 @@ def test_pattern_noise_in_only_one_quadrant(set_random_seed):
...
+
+def test_compute_snr():
+    data = np.random.normal(1000.0, 20.0, size=(200, 100))
+    snr = pattern_noise.compute_snr(data)
+    assert len(snr) == data.shape[1]-1
+    assert all(snr < 5)
+
+
+def test_get_odd_integer():
+    assert pattern_noise.get_odd_integer(1.5) == 3
```
cmccully: Shouldn't this go to 1?

cmccully: i.e. the closest odd integer to 1.5 is 1.

banzai/qc/pattern_noise.py
```diff
@@ -21,49 +22,197 @@ def group_by_keywords(self):
...
+    Returns
+    -------
+    snr_convolved : numpy array
+        Array of size nwavelets x len(snr) of snr convolved with different wavelet widths
+    """
+
+    widths = np.linspace(len(snr) * fractional_wavelet_width_min,
+                         len(snr) * fractional_wavelet_width_max,
+                         nwavelets)
+    return cwt(snr, ricker, widths)
```
cmccully: This might be more readable as scipy.signal.cwt

banzai/main.py
turnerm: I think we decided to run the pattern noise checker on the reduced data, just to minimize occurrences of FFT artifacts due to e.g. non-flat field etc.

banzai/qc/pattern_noise.py
turnerm: I changed it to `PEAK_POSITION_STD_THRESHOLD`

banzai/qc/pattern_noise.py
turnerm: Nope, good point, I'll take this out.

banzai/qc/pattern_noise.py
turnerm: Changed

banzai/tests/test_pattern_noise_qc.py
turnerm: I guess this function really finds the odd integer ceiling. I have updated the docstring, I can also modify the function to really find the closest odd integer but I think that may be overkill.

banzai/qc/pattern_noise.py
turnerm: Is there a case you have in mind where this might happen? `peak_maxima` and `std_maxima` come from `np.max` and `np.std` which are pretty robust. I added a test where the image is made of NaNs and even then `check_for_pattern_noise` runs all the way through without a problem.

PR #119, opened 2018-07-27
hotfix/missing_telescope_exception

gmbrandt:
Fixed read_images and get_telescope_id so that they do not crash if the telescope is not in the database.
The LCO database has NRES frames under TELESCOP id, not instrument. Thus whenever an instance of the Image class from images.py is created, using an NRES pipeline context, then we get an exception from get_telescope_id and so the Image instance is not created. 
Also fixed read_images so that it would not return an empty list if the fits file which was imported already had a BPM.

### DIFF
--- a/banzai/dbs.py
+++ b/banzai/dbs.py
@@ -293,7 +293,6 @@ class TelescopeMissingException(Exception):
 
 
 def get_telescope_id(site, instrument, db_address=_DEFAULT_DB):
-    # TODO:  This dies if the telescope is not in the telescopes table. Maybe ping the configdb?
     db_session = get_session(db_address=db_address)
     criteria = (Telescope.site == site) & (Telescope.instrument == instrument)
     telescope = db_session.query(Telescope).filter(criteria).first()
@@ -301,8 +300,11 @@ def get_telescope_id(site, instrument, db_address=_DEFAULT_DB):
     if telescope is None:
         err_msg = '{site}/{instrument} is not in the database.'.format(site=site,
                                                                        instrument=instrument)
-        raise TelescopeMissingException(err_msg)
-    return telescope.id
+        logger.error(err_msg, extra={'tags': {'site': site, 'instrument': instrument}})
+        telescope_id = None
+    else:
+        telescope_id = telescope.id
+    return telescope_id
 
 
 def get_telescope(telescope_id, db_address=_DEFAULT_DB):
--- a/banzai/images.py
+++ b/banzai/images.py
@@ -49,7 +49,7 @@ class Image(object):
                 self.telescope_id = dbs.get_telescope_id(self.site, header.get('TELESCOP'),
                                                          db_address=pipeline_context.db_address)
                 self.instrument = header.get('TELESCOP')
-                
+
         self.epoch = str(header.get('DAY-OBS'))
         self.nx = header.get('NAXIS1')
         self.ny = header.get('NAXIS2')
--- a/banzai/images.py
+++ b/banzai/images.py
@@ -172,6 +172,8 @@ def read_images(image_list, pipeline_context):
     for filename in image_list:
         try:
             image = Image(pipeline_context, filename=filename)
+            if image.telescope_id is None:
+                raise dbs.TelescopeMissingException
             munge(image, pipeline_context)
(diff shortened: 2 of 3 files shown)


banzai/dbs.py
```diff
@@ -293,16 +293,18 @@ class TelescopeMissingException(Exception):
```
turnerm: Since `get_telescope_id` can now return `None`,  `populate_bpm_table` might try to `add_or_update_record` with `None` in the `telescope_id` field. Could this be a problem?

banzai/images.py
```diff
@@ -172,6 +172,8 @@ def read_images(image_list, pipeline_context):
     for filename in image_list:
         try:
             image = Image(pipeline_context, filename=filename)
```
turnerm: Can probably remove `if self.site is not None and self.instrument is not None:` from line 45 of `images.py`

banzai/images.py
```diff
@@ -172,6 +172,8 @@ def read_images(image_list, pipeline_context):
...
+            if image.telescope_id is None:
+                raise dbs.TelescopeMissingException
```
turnerm: Maybe add an error message to this

banzai/images.py
```diff
@@ -172,6 +172,8 @@ def read_images(image_list, pipeline_context):
...
             munge(image, pipeline_context)
             if image.bpm is None:
                 bpm = image_utils.get_bpm(image, pipeline_context)
```
turnerm: Could this line be moved to `Image` class instantiator? It could help simplify the logic here.

banzai/dbs.py
cmccully: Yes. We should make sure that doesn't happen. Good catch.

banzai/images.py
cmccully: It would. But that totally defaults the purpose of this. The problem is that right now we can't create Image objects without a DB even for testing.

turnerm: Ah yes I had misunderstood how the BPM was set in the instantiator. At any rate I guess this section will be changed in the "override BPM" branch.

banzai/images.py
gmbrandt: I think we want that there so that it does not search the database if instrument or site are None. If so, a search through the db will raise and exception and any instance of Image where the site and instrument do not correspond to a telescope will fail to initilize

banzai/dbs.py
gmbrandt: Curtis, because we have to do different things to NRES frames and image frames (e.g. trace fitting), why don't we add an attribute to the pipline context, e.g. is_nres_frame? It seems like we will need something like that in the future anyway, and then here we could just have 
```
if pipeline_context.is_nres_frame:
    self.instrument = header.get('TELESCOP')
else:
    self.instrument=  header.get('INSTRUME')
```
 right before the if self.site is not None etc check. Then we could revert back to having get_telescope_id just throwing an exception, and not returning none.

cmccully: I think I would just say, if telescope_id is None, try to set the instrument to Telescop.

banzai/images.py
cmccully: Done in another PR.

banzai/images.py
gmbrandt: I agree. turnerm further pointed out that the entire block can be removed, i.e. 45, 48 and 49, now that get_telescope_id returns None now.

banzai/dbs.py
gmbrandt: Ah yes, I agree. that is nice.

gmbrandt: Is that in the next PR or should I add it into this one?

gmbrandt:
This makes it so that get_telescope_id returns None instead of crashing if it does not find the telescope. In read images, if we do not initially find a telescope, we search for one using the TELESCOP name instead of instrument name, which is how NRES frames are indexed. 

I added a check inside of munge.munge which checks for CRPIX 1 and 2, which adds the headers to the image object if they do not exist. NRES frames lack these, and thus do not pass through the pipe.

Thus NRES frames get properly piped through now.

cmccully:
Fixed via #120

PR #120, opened 2018-07-30
Added override to bad pixel mask requirement.

cmccully:


banzai/main.py
```diff
@@ -233,6 +234,8 @@ def reduce_night():
                         help='ElasticSearch index to use for QC results')
     parser.add_argument('--es-doc-type', dest='elasticsearch_doc_type', default='qc',
                         help='Elasticsearch document type for QC records')
+    parser.add_argument('--bpm-not-required', dest='bpm_required', default=True, action='store_false',
+                        help='Do not require a bad pixel to be present to reduce data (Default BPM contains all zeros)')
```
gmbrandt: Might be more clear to have --bpm-required with a default of False. The double negative confuses me a bit, e.g that I have to set bpm-not-required to False in order to have banzai fetch a bpm. Or is this supposed to be --bpm-required? I am a bit confused.

cmccully:
For the NRES project and for testing, we need the ability to create Image objects without having a connection to the DB. This moves the exceptions around to only raise an exception for production systems. We also add a flag to the context object to override the requirement of a bad pixel mask (we make a default bad pixel mask of all zeros) so we can easily reduce test data before a bad pixel mask exists.

banzai/dbs.py
```diff
@@ -292,25 +297,41 @@ class TelescopeMissingException(Exception):
...
+    site = header.get('SITEID')
+    instrument = header.get('INSTRUME')
+    telescope = query_for_telescope(db_address, site, instrument)
+    if telescope is None:
+        telescope = query_for_telescope(db_address, site, header.get('TELESCOPE'))
     if telescope is None:
-        raise TelescopeMissingException('Telescope ID {id} is not in the database.'.format(id=telescope_id))
+        err_msg = '{site}/{instrument} is not in the database.'.format(site=site, instrument=instrument)
+        logger.error(err_msg, extra={'tags': {'site': site, 'instrument': instrument}})
     return telescope
```
turnerm: What if you raise a `TelescopeMissingException` in `query_for_telescope`, and then use try/except in `get_telescope` instead. I guess the downside is that the error will pop up every time you give it an NRES frame. But something like this:

```
def query_for_telescope(db_address, site, instrument):
    db_session = get_session(db_address=db_address)
    criteria = (Telescope.site == site) & (Telescope.instrument == instrument)
    telescope = db_session.query(Telescope).filter(criteria).first()
    db_session.close()
    if telescope is None:
        err_msg = '{site}/{instrument} is not in the database.'.format(site=site, instrument=instrument)
        raise TelescopeMissingException(err_msg)
    return telescope


def get_telescope(header, db_address=_DEFAULT_DB):
    site = header.get('SITEID')
    instrument = header.get('INSTRUME')
    try:
        telescope = query_for_telescope(db_address, site, instrument)
    except TelescopeMissingException:
        telescope = query_for_telescope(db_address, site, header.get('TELESCOPE'))
    except TelescopeMissingException:
        telescope = None
    return telescope

```

banzai/images.py
```diff
@@ -40,13 +40,11 @@ def __init__(self, pipeline_context, filename=None, data=None, header=None,
...
-        if self.site is not None and self.instrument is not None:
-            self.telescope_id = dbs.get_telescope_id(self.site, self.instrument,
-                                                     db_address=pipeline_context.db_address)
-        else:
-            self.telescope_id = None
+        self.telescope = dbs.get_telescope(self.header)
+        if self.telescope is not None:
+            self.site = self.telescope.site
+            self.instrument = self.telescope.instrument
+
```
turnerm: Would it make sense to fall back to header values for `site` and `instrument` if a telescope isn't found?

banzai/utils/image_utils.py
```diff
@@ -106,36 +110,44 @@ def save_images(pipeline_context, images, master_calibration=False):
...
-        image.header['L1IDMASK'] = ('', 'Id. of mask file used')
+    if pipeline_context.bpm_required and bpm_filename is None:
+        raise MissingBPMError('No Bad Pixel Mask file exists for this image.')
+    elif bpm_filename is None:
+        load_default_bpm(image)
+    else:
+        load_bpm_file(bpm_filename, image)
+
+
+def load_default_bpm(image):
```
turnerm: Maybe call this "null" instead of "default"?

banzai/utils/image_utils.py
```diff
@@ -106,36 +110,44 @@ def save_images(pipeline_context, images, master_calibration=False):
...
-            raise ValueError(err_msg)
-        image.header['L1IDMASK'] = (os.path.basename(bpm_filename), 'Id. of mask file used')
-
-    return bpm
+        image.bpm = np.array(bpm_hdu[0].data, dtype=np.uint8)
+    if not bpm_has_valid_size(image.bpm, image):
+        tags = logs.image_config_to_tags(image, None)
+        logs.add_tag(tags, 'filename', image.filename)
+        logger.error('BPM shape mismatch', extra=tags)
+        raise ValueError('BPM shape mismatch')
```
turnerm: In case of a shape mismatch, would you want to fall back on to a null BPM?

banzai/utils/image_utils.py
gmbrandt: I agree

banzai/utils/image_utils.py
gmbrandt: I think this only gets called if bpm_required is false and bpm_filename is not none - in which case it is trying to do image.bpm = bpm (from the bpm_filename).  So if there is a shape mismatch I think it would be good to throw an error and not load the bpm, which would stop the images from being processed later on I believe? Just so that we really know when we have configured the bpm's incorrectly

banzai/images.py
gmbrandt: I agree, good idea. how about something like
self.site = header.get('SITEID')
self.instrument = header.get('instrument')
prior to the if self.telescope is not None call

banzai/dbs.py
```diff
@@ -292,25 +297,41 @@ class TelescopeMissingException(Exception):
     pass
 
 
-def get_telescope_id(site, instrument, db_address=_DEFAULT_DB):
-    # TODO:  This dies if the telescope is not in the telescopes table. Maybe ping the configdb?
+def get_telescope_id(header, db_address=_DEFAULT_DB):
     db_session = get_session(db_address=db_address)
     criteria = (Telescope.site == site) & (Telescope.instrument == instrument)
```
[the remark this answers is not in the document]
gmbrandt: yes, why did this not throw an error? This function just needs lines 327 and 328 from get_telescope below.

banzai/dbs.py
gmbrandt: I like this

gmbrandt: Although we still want to set self.instrument = header.get('TELESCOP')
for the image object after searching the telescope, so that the designations are clear down the line

gmbrandt: IS there some way we could get it to not throw an error for NRES frames every time?

banzai/dbs.py
```diff
@@ -292,25 +297,26 @@ class TelescopeMissingException(Exception):
...
-def get_telescope(telescope_id, db_address=_DEFAULT_DB):
-    db_session = get_session(db_address=db_address)
-    telescope = db_session.query(Telescope).filter(Telescope.id == telescope_id).first()
-    db_session.close()
+def get_telescope(header, db_address=_DEFAULT_DB):
+    site = header.get('SITEID')
+    instrument = header.get('INSTRUME')
+    telescope_for_instrument = _query_for_telescope(db_address, site, instrument)
+    telescope = telescope_for_instrument if telescope_for_instrument is not None \
+        else _query_for_telescope(db_address, site, header.get('TELESCOP'))
```
gmbrandt: This is very clear now.

banzai/images.py
```diff
@@ -40,13 +40,14 @@ def __init__(self, pipeline_context, filename=None, data=None, header=None,
 
         self.request_number = header.get('REQNUM')
 
-        self.site = header.get('SITEID')
-        self.instrument = header.get('INSTRUME')
-        if self.site is not None and self.instrument is not None:
-            self.telescope_id = dbs.get_telescope_id(self.site, self.instrument,
-                                                     db_address=pipeline_context.db_address)
+        self.telescope = dbs.get_telescope(self.header)
```
turnerm: I think you need to pass `db_address` here

banzai/utils/image_utils.py
```diff
@@ -106,36 +110,44 @@ def save_images(pipeline_context, images, master_calibration=False):
     return output_files
 
 
-def get_bpm(image, pipeline_context):
+def load_bpm(image, pipeline_context):
     bpm_filename = dbs.get_bpm(image.telescope_id, image.ccdsum,
```
turnerm: I think this should be `image.telescope.id`

turnerm:
Also, line 465 of `dbs.py`, in the method `get_master_calibration_image`, you should update to `image.telescope.id` (instead of `image.telescope_id`)

banzai/utils/image_utils.py
gmbrandt: I agree

cmccully:
@gmbrandt I think the mysqldb issue is not a bug here.

banzai/munge.py
```diff
@@ -12,21 +12,18 @@ class SinistroModeNotSupported(Exception):
...
 def munge(image, pipeline_context):
-    telescope = dbs.get_telescope(image.telescope_id,
-                                  db_address=pipeline_context.db_address)
-
-    if 'sinistro' in telescope.camera_type.lower():
+    if 'sinistro' in image.telescope.camera_type.lower():
         if sinistro_mode_is_supported(image):
             munge_sinistro(image)
         else:
             raise SinistroModeNotSupported('Sinistro mode not supported {f}'.format(f=image.filename))
```
gmbrandt: Can we add here the NRES CRPIX1 & 2  fix? Maybe it is more appropriate in a second PR after this one,
but the following is required to get NRES frames through trim.trimmer.
```
    if image.header.get('CRPIX1') is None:
        image.header['CRPIX1'] = 0
    if image.header.get('CRPIX2') is None:
image.header['CRPIX2'] = 0
```

banzai/images.py
```diff
@@ -40,13 +40,15 @@ def __init__(self, pipeline_context, filename=None, data=None, header=None,
...
         self.request_number = header.get('REQNUM')
 
-        self.site = header.get('SITEID')
-        self.instrument = header.get('INSTRUME')
-        if self.site is not None and self.instrument is not None:
-            self.telescope_id = dbs.get_telescope_id(self.site, self.instrument,
-                                                     db_address=pipeline_context.db_address)
-        else:
-            self.telescope_id = None
+        if len(self.header) > 0:
```
gmbrandt: Is this line to correctly pass FakeImages?

cmccully: Fixed.

PR #121, opened 2018-08-21
End to end tests

cmccully:
Added end to end tests for BANZAI.

### DIFF
--- a/banzai/tests/test_end_to_end.py
+++ b/banzai/tests/test_end_to_end.py
@@ -1,0 +1,92 @@
+import pytest
+import os
+from glob import glob
+
+data_root = '/archive/engineering'
+
+sites = [os.path.basename(site_path) for site_path in glob(os.path.join(data_root, '*'))]
+instruments = [os.path.join(site, os.path.basename(instrument_path)) for site in sites
+               for instrument_path in glob(os.path.join(os.path.join(data_root, site, '*')))]
+
+days_obs = [os.path.join(instrument, os.path.basename(dayobs_path)) for instrument in instruments
+            for dayobs_path in glob(os.path.join(data_root, instrument, '*'))]
+
+
+def run_banzai(entry_point):
+    for day_obs in days_obs:
+        raw_path = os.path.join(data_root, day_obs, 'raw')
+        command = '{cmd} --raw-path {raw_path} --fpack --db-address={db_address}'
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    labels:
+      io.rancher.container.pull_image: always
+      io.rancher.sidekicks: BANZAIpipelineE2EData,AstrometryCatalogs
+    logging:
+      options:
+        max-size: '100m'
+        max-file: '3'
+volumes:
+  banzaie2e:
+    driver: local
```
kurczynski: I think it sets the driver to local by default.

docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    volumes:
+    - astrometry:/opt/astrometry.net/data
+    logging:
+      options:
+        max-size: '100m'
+        max-file: '3'
+  BANZAITest:
+    image: ${DOCKER_IMG}
+    entrypoint:
+    - tail
```
kurczynski: Why is this entrypoint `tail`?

docker-compose.yml
cmccully: I believe I had to set this for Rancher or it wouldn't properly clean up the volume when the service was removed.

docker-compose.yml
cmccully: This is just a dummy process to keep the container running so that we can exec commands in the container. I'm open to suggestions on that though.

banzai/tests/test_end_to_end.py
```diff
@@ -0,0 +1,136 @@
...
+        if calibration_type.lower() == 'skyflat':
+            # Group by filter
+            observed_filters = []
+            for raw_filename in raw_filenames_for_this_dayobs:
+                skyflat_hdu = fits_utils.open_fits_file(raw_filename)
+                observed_filters.append(skyflat_hdu[0].header['FILTER'])
+            observed_filters = set(observed_filters)
+            number_of_stacks_that_should_have_been_created += len(observed_filters)
+        else:
+            # Just one calibration per night
```
turnerm: Should we also check binning? Maybe that's for the future...

cmccully: I think that is a future enhancement.

PR #122, opened 2018-08-22
Enabled rejection of bias and dark frames when creating a master.

cmccully:


banzai/tests/test_bias_maker.py
```diff
@@ -27,28 +28,6 @@ def test_group_by_keywords():
...
-@mock.patch('banzai.bias.Image')
-def test_header_master_bias_level_returns_2(mock_image):
-    maker = BiasMaker(FakeContext())
-
-    maker.do_stage([FakeBiasImage(image_multiplier=2.0) for x in range(6)])
-
-    args, kwargs = mock_image.call_args
-    header = kwargs['header']
-    assert header['BIASLVL'] == 2.0
-
```
turnerm: Maybe include a check for the final `BIASVAL` value, but one that reflects that it is now calculated as the mean of the values from the input file headers.

cmccully: Done.

PR #124, opened 2018-08-23
Took boolean out of add_tag

turnerm:


banzai/stages.py
```diff
@@ -211,16 +211,19 @@ def apply_master_calibration(self, images, master_calibration_image, logging_tag
...
-                          "master_comparison.failed": frame_is_bad,
                           "master_comparison.comparison_master_filename": master_calibration_image.filename}
 
             for qc_check, qc_result in qc_results.items():
                 logs.add_tag(logging_tags, qc_check, qc_result)
             logs.add_tag(logging_tags, 'filename', image.filename)
             logs.add_tag(logging_tags, 'master_comparison_filename', master_calibration_image.filename)
+
             self.logger.info("Performing comparison to last good master {caltype} frame".format(
                 caltype=self.calibration_type), extra=logging_tags)
```
cmccully: While you are here, this is weird spacing. It would be better to define a message variable and then call format in a separate line.

banzai/dark.py
```diff
@@ -124,6 +124,7 @@ def reject_images(self):
         return True
 
     def noise_model(self, image):
-        noise = (image.readnoise ** 2.0 + np.abs(image.data * image.exptime)) ** 0.5
+        poisson_noise = np.where(image.data > 0, image.data * image.exptime, 0)
```
cmccully: Why is this not 0.0?

PR #128, opened 2018-09-04
Fix/pattern noise snr thresh

turnerm:


banzai/qc/pattern_noise.py
```diff
@@ -58,13 +69,32 @@ def check_for_pattern_noise(self, data):
         trimmed_data = trim_image_edges(data)
         power_2d = get_2d_power_band(trimmed_data)
         snr = compute_snr(power_2d)
-        convolved_snr = convolve_snr_with_wavelet(snr)
-        peak_maxima, std_maxima = get_peak_parameters(convolved_snr)
+        fraction_pixels_above_threshold = self.get_n_grouped_pixels_above_threshold(snr) / len(snr)
```
cmccully: In Python 2, this will do integer division.

banzai/qc/pattern_noise.py
```diff
@@ -58,13 +69,32 @@ def check_for_pattern_noise(self, data):
...
+        -------
+        n_grouped_pixels_above_threshold : numpy array
+            The number of SNR pixels with values above SNR_THRESHOLD
+            that are in groups of at least MIN_ADJACENT_PIXELS
+        """
+
+        idx_above_thresh = np.where(snr > self.SNR_THRESHOLD)[0]
+        consecutive_group_lengths = np.array([len(list(map(itemgetter(1), g))) for k, g in
+                                              groupby(enumerate(idx_above_thresh), key=lambda x: x[0]-x[1])])
+        n_grouped_pixels_above_threshold = sum(consecutive_group_lengths[consecutive_group_lengths
```
cmccully: Might be worth splitting this into two lines.
maybe define the criteria on one line and then the do the sum on the next?

banzai/qc/pattern_noise.py
turnerm: I've updated to cast the denominator to a float.

banzai/qc/pattern_noise.py
turnerm: Done

PR #129, opened 2018-09-06
Filter NRES and FLOYDS data

cmccully:
Added the ability to filter out non-imaging data from banzai pipeline. This allows us to run a banzai-nres pipeline in parallel.


banzai/preview.py
```diff
@@ -0,0 +1,76 @@
+from banzai import dbs
+from banzai.utils import file_utils
+from banzai import logs
+from banzai.utils.image_utils import image_passes_criteria
+
+logger = logs.get_logger(__name__)
+
+
```
turnerm: Should `PreviewModeListener` be moved here?

banzai/main.py
```diff
@@ -45,6 +49,10 @@
                   astrometry.WCSSolver,
```
turnerm: `ordered_stages` should probably be all caps too. Your favourite!

cmccully: You are of course correct.

banzai/preview.py
cmccully: Indeed.

cmccully: Oh. Right. Circular imports make this really complicated. This will be refactored once we merge end of night and preview modes.

PR #131, opened 2018-09-13
Fixed how filters for telescope classes were plumbed.

cmccully:


banzai/main.py
```diff
@@ -86,8 +86,8 @@ def get_stages_todo(last_stage=None, extra_stages=None):
     return stages_todo
 
 
-def run_end_of_night_from_console(scripts_to_run):
-    pipeline_context = parse_end_of_night_command_line_arguments()
+def run_end_of_night_from_console(scripts_to_run, selection_criteria):
```
turnerm: Could you set `selection_criteria=IMAGING_CRITERIA` as a default kwarg, to avoid having to pass it in all the calls below?

cmccully: This will be addressed more in a future PR.

PR #132, opened 2018-09-13
Broke headers out to an object

kurczynski:
It seems like there is an `Image` object that is comprised of other objects such as a `Header` object.

From a high level, everything that is needed to create an `Image` should be given to the `Image` constructor (would be nice if it was just a filename of a file to read from), and then instance objects of `Image` can be populated with that data inside `Image`.

### DIFF
--- a/banzai/image/header.py
+++ b/banzai/image/header.py
@@ -1,0 +1,25 @@
+from banzai import logs
+from banzai.utils import date_utils, fits_utils
+
+logger = logs.get_logger(__name__)
+
+class Header(object):
+    def __init__(self, header: dict):
+        self.request_number = header['REQNUM']
+        self.epoch = str(header['DAY-OBS'])
+        self.nx = header['NAXIS1']
+        self.ny = header['NAXIS2']
+        self.block_id = header['BLKUID']
+        self.molecule_id = header['MOLUID']
+        self.ccdsum = header['CCDSUM']
+        self.filter = header['FILTER']
+        self.obstype = header['OBSTYPE']
+        self.gain = eval(str(header['GAIN']))
+        self.site = header['SITEID']
+        self.instrument = header['INSTRUME']
+
+        self.exptime = float(header.get('EXPTIME', 0.0))
+        self.dateobs = date_utils.parse_date_obs(header.get('DATE-OBS', '1900-01-01T00:00:00.00000'))
+        self.readnoise = float(header.get('RDNOISE', 0.0))
+        self.ra, self.dec = fits_utils.parse_ra_dec(header)
+        self.pixel_scale = float(header.get('PIXSCALE', 0.0))
--- a/banzai/image/image.py
+++ b/banzai/image/image.py
@@ -7,7 +7,7 @@ import tempfile
 import shutil
 
 from banzai import dbs
-from banzai.utils import date_utils
+from banzai.image.header import Header
 from banzai.utils import fits_utils
 from banzai.utils import image_utils
 from banzai import logs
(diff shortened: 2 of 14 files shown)


banzai/image/header.py
```diff
@@ -0,0 +1,25 @@
+from banzai import logs
+from banzai.utils import date_utils, fits_utils
+
+logger = logs.get_logger(__name__)
+
+class Header(object):
+    def __init__(self, header: dict):
+        self.request_number = header['REQNUM']
```
kurczynski: I'm assuming that these values will always be in the header. If not, a `KeyError` will be raised which can be wrapped with something more descriptive.

If these values will not always be in the header, they should have defaults like at the end of this constructor.

banzai/image/image.py
```diff
@@ -18,60 +18,42 @@
 
 class Image(object):
 
-    def __init__(self, pipeline_context, filename=None, data=None, header=None,
-                 extension_headers=None, bpm=None):
-        if header is None:
-            header = {}
-
+    def __init__(self, pipeline_context, filename: str = None, data=None, header: Header = None,
```
kurczynski: When an `Image` object is created, where does the `header` come from? Is it not contained in the file `filename` that is being passed in here?

banzai/image/image.py
```diff
@@ -18,60 +18,42 @@
...
-        self.exptime = float(header.get('EXPTIME', 0.0))
-        self.dateobs = date_utils.parse_date_obs(header.get('DATE-OBS', '1900-01-01T00:00:00.00000'))
-        self.readnoise = float(header.get('RDNOISE', 0.0))
-        self.ra, self.dec = fits_utils.parse_ra_dec(header)
-        self.pixel_scale = float(header.get('PIXSCALE', 0.0))
-        self.catalog = None
+        return gain
 
-    def _init_telescope_info(self, pipeline_context):
-        if len(self.header) > 0:
```
kurczynski: When would a header ever have a zero length?

banzai/image/image.py
```diff
@@ -138,11 +120,11 @@ def write_catalog(self, filename, nsources=None):
     def add_history(self, msg):
         self.header.add_history(msg)
 
-    def data_is_3d(self):
+    def is_data_3d(self) -> bool:
```
kurczynski: Prefixing `bool` variables and methods with `is` is common.

banzai/image/image.py
```diff
@@ -18,60 +18,42 @@
...
 class Image(object):
 
-    def __init__(self, pipeline_context, filename=None, data=None, header=None,
-                 extension_headers=None, bpm=None):
-        if header is None:
-            header = {}
-
+    def __init__(self, pipeline_context, filename: str = None, data=None, header: Header = None,
+                 extension_headers: list = None, bpm=None):
         if extension_headers is None:
```
kurczynski: `extension_header` might be a good candidate for an object as well.

banzai/image/image.py
cmccully: This could be factory class methods.

banzai/image/image.py
```diff
@@ -18,60 +18,42 @@
...
             extension_headers = []
 
         if filename is not None:
             data, header, bpm, extension_headers = fits_utils.open_image(filename)
-            if '.fz' == filename[-3:]:
-                filename = filename[:-3]
+            filename = os.path.splitext(filename)[0]
```
cmccully: filename = filename.replace('.fits.fz', '.fits')

banzai/image/image.py
```diff
@@ -18,60 +18,42 @@
...
+
             self.filename = os.path.basename(filename)
 
         self.data = data
-        self.header = header
+        self.header = Header(header)
```
cmccully: fits.Header()

banzai/image/image.py
cmccully: This would be if it defaulted to an empty dict. We could do `if not self.header:` if that is better.

banzai/image/image.py
```diff
@@ -18,60 +18,42 @@
...
-            telescope = dbs.get_telescope(self.header, db_address=pipeline_context.db_address)
+    def _init_telescope_info(self, pipeline_context, header):
+        if header is not None:
+            telescope = dbs.get_telescope(header, db_address=pipeline_context.db_address)
             if telescope is not None:
                 site = telescope.site
                 instrument = telescope.instrument
             else:
-                site = self.header.get('SITEID')
-                instrument = self.header.get('INSTRUME')
```
cmccully: I think it is just better to remove this else entirely.

cmccully:
Much of this has been implemented in other PRs. We will try to keep this in mind moving forward.

PR #133, opened 2018-09-14
Feature/images support astropy tables

gmbrandt:
This allows images to save astropy data tables which are added to an instance of the class. This is applicable for NRES frames which will save trace related files as astropy tables.


banzai/images.py
```diff
@@ -96,24 +98,25 @@ def writeto(self, filename, fpack=False):
             bpm_hdu = fits.ImageHDU(self.bpm.astype(np.uint8))
             bpm_hdu.name = 'BPM'
             hdu_list.append(bpm_hdu)
+        hdu_list = self.add_astropy_data_tables_to_hdu_list_to_be_saved(hdu_list)
```
turnerm: For consistency, you could also refactor the catalog and bpm additions to their own methods

banzai/images.py
```diff
@@ -96,24 +98,25 @@ def writeto(self, filename, fpack=False):
...
 
-        hdu_list = fits.HDUList(hdu_list)
+        fits_hdu_list = fits.HDUList(hdu_list)
```
turnerm: Why the variable name change?

banzai/images.py
```diff
@@ -125,6 +128,19 @@ def writeto(self, filename, fpack=False):
                 self.filename += '.fz'
             shutil.move(os.path.join(temp_directory, base_filename), filename)
 
+    def regenerate_astropy_data_table_from_fits_table_hdu(self, hdu_list, table_extension_name):
```
turnerm: If this is just for testing, then it should go in `test_image.py`

banzai/images.py
gmbrandt: will do

banzai/images.py
gmbrandt: because hdu_list was both used for a list of hdu objects, and as the output of fits.HDUList(list_of_hdus). I thought that was confusing.

banzai/images.py
gmbrandt: it is for reorganizing tables into proper astropy tables once they have been loaded from disc

turnerm: Actually, it can go outside of the Image class in `image.py`.

banzai/images.py
```diff
@@ -125,6 +128,19 @@ def writeto(self, filename, fpack=False):
...
+        fits_rec_table = hdu_list[table_extension_name].data
+        astropy_table = Table(fits_rec_table, meta={'name': table_extension_name})
+        return astropy_table
+
+    def add_astropy_data_tables_to_hdu_list_to_be_saved(self, hdu_list):
```
turnerm: Maybe make this a private method?

banzai/images.py
gmbrandt: is this because it is 'static'?

banzai/images.py
cmccully: fits is a file type, so I don't think it is the best variable name here.

banzai/images.py
cmccully: I would prefer to refactor to just data table. In principle this could be a pandas dataframe, etc. Nothing we do here is astropy specific (I think).

banzai/images.py
turnerm: Yep

banzai/images.py
turnerm: *protected, sorry

banzai/images.py
```diff
@@ -16,10 +17,26 @@
...
+
+    def __setitem__(self, key, value):
+        self._data_table[key] = value
+
+
 class Image(object):
 
-    def __init__(self, pipeline_context, filename=None, data=None, header=None,
-                 extension_headers=None, bpm=None):
+    def __init__(self, pipeline_context, filename=None, data=None, data_tables={},
```
turnerm: You should set `data_tables` to `None` and then to `{}` if it is `None`, as done with `header`

banzai/images.py
```diff
@@ -88,32 +105,26 @@ def writeto(self, filename, fpack=False):
...
-            hdu_list.append(table_hdu)
-        if self.bpm is not None:
-            bpm_hdu = fits.ImageHDU(self.bpm.astype(np.uint8))
-            bpm_hdu.name = 'BPM'
-            hdu_list.append(bpm_hdu)
+        hdu_list = self.add_bpm_to_hdu_list(hdu_list)
+        hdu_list = self.add_data_tables_to_hdu_list(hdu_list)
 
-        hdu_list = fits.HDUList(hdu_list)
+        fits_hdu_list = fits.HDUList(hdu_list)
```
turnerm: As Curtis mentioned, maybe this needs to be refactored a bit (and also I'm not sure he wanted a separate BPM adding method). Perhaps lines 107 to 123 could be moved to their own method. And then you can just go back to calling this variable `hdu_list`.

banzai/images.py
```diff
@@ -125,15 +136,33 @@ def writeto(self, filename, fpack=False):
...
+        :param hdu_list: a list of hdu objects.
+        :return: a list of hdu objects with a FitsBinTableHDU added
+        """
+        for key in self.data_tables:
+            table_hdu = fits_utils.table_to_fits(self.data_tables[key]._data_table)
+            table_hdu.name = self.data_tables[key].name
+            hdu_list.append(table_hdu)
+        return hdu_list
+
+    def add_bpm_to_hdu_list(self, hdu_list):
```
turnerm: Not sure if / how this will be refactored, but in its current state it should probably be made protected, and come before `add_data_tables`, i.e. be consistent with the order that the methods are called.

banzai/images.py
```diff
@@ -191,3 +220,16 @@ def read_images(image_list, pipeline_context):
             logger.error('Error loading image: {error}'.format(error=e), extra={'tags': {'filename': filename}})
             continue
     return images
+
+
+def regenerate_data_table_from_fits_hdu_list(hdu_list, table_extension_name, input_dictionary={}):
```
turnerm: Same comment about `{}`

banzai/images.py
```diff
@@ -88,32 +105,26 @@ def writeto(self, filename, fpack=False):
...
-        if self.catalog is not None:
-            table_hdu = fits_utils.table_to_fits(self.catalog)
-            table_hdu.name = 'CAT'
-            hdu_list.append(table_hdu)
-        if self.bpm is not None:
-            bpm_hdu = fits.ImageHDU(self.bpm.astype(np.uint8))
-            bpm_hdu.name = 'BPM'
-            hdu_list.append(bpm_hdu)
+        hdu_list = self.add_bpm_to_hdu_list(hdu_list)
+        hdu_list = self.add_data_tables_to_hdu_list(hdu_list)
```
turnerm: This order will change the structure of the output images, i.e. now HDU 1 will be the BPM and HDU 2 the catalogue. Maybe Curtis can comment on whether that matters though.

gmbrandt: Good catch. If that matters @cmccully , I will revert to the old order

banzai/images.py
gmbrandt: the catalog business is necessary to get banzai-nres working so it can move forward again. I would like to leave refactoring of the write_to function after this HDUList thing for another PR. What do you think? For now, I will change the variable back to hdu_list.

banzai/images.py
gmbrandt: I will protect both these methods, thanks

banzai/images.py
gmbrandt: fixed

banzai/images.py
gmbrandt: fixed

banzai/images.py
```diff
@@ -16,10 +17,26 @@
...
 
 
+class DataTable(object):
+    """
+    Object for storing astropy (or another table type) tables with an additional .name attribute which
+    determines the tables' extension name when it is saved as a fits file.
+    """
+    def __init__(self, data_table, name):
+        self.name = name
+        self._data_table = data_table
```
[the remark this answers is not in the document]
gmbrandt: @cmccully what was the reason for protecting this variable? I do not remember

PR #136, opened 2018-10-04
Feature/logging update

turnerm:
The main structural changes are in `logs.py`, most everything else is just plumbing. 

Since I had to modify the imports in most files, I took it upon myself to PEP8-ify things in the process. 

Currently many tests are failing, but that's because I need to update the tests themselves. E2E on actual (non-sinistro...) data should work.

### DIFF
--- a/banzai/astrometry.py
+++ b/banzai/astrometry.py
@@ -1,5 +1,3 @@
-__author__ = 'cmccully'
-
 import os
 import subprocess
 import shlex
--- a/banzai/astrometry.py
+++ b/banzai/astrometry.py
@@ -1,16 +1,21 @@
-from __future__ import absolute_import, division, print_function, unicode_literals
-from banzai.stages import Stage
-from banzai import logs
-from banzai.utils import image_utils
-import os, subprocess, shlex
-from astropy.io import fits
+__author__ = 'cmccully'
+
+import os
+import subprocess
+import shlex
 import tempfile
+import logging
+
+from astropy.io import fits
 from astropy.wcs import WCS
 from astropy.coordinates import SkyCoord
 from astropy import units
 import numpy as np
 
-__author__ = 'cmccully'
+from banzai.stages import Stage
+from banzai.utils import image_utils
+
+logger = logging.getLogger(__name__)
 
 
 class WCSSolver(Stage):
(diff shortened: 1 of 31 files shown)


turnerm:
Should be ready for review now.

banzai/astrometry.py
```diff
@@ -1,16 +1,21 @@
-from __future__ import absolute_import, division, print_function, unicode_literals
-from banzai.stages import Stage
-from banzai import logs
-from banzai.utils import image_utils
-import os, subprocess, shlex
-from astropy.io import fits
+__author__ = 'cmccully'
```
cmccully: We can remove the author line.

banzai/bias.py
```diff
@@ -1,16 +1,15 @@
-from __future__ import absolute_import, division, print_function, unicode_literals
+__author__ = 'cmccully'
```
cmccully: We can remove this line too.

banzai/bias.py
```diff
@@ -113,19 +109,15 @@ def group_by_keywords(self):
...
-            logs.add_tag(logging_tags, 'filename', os.path.basename(image.filename))
-
             # Subtract the overscan if it exists
             if image.data_is_3d():
                 for i in range(image.get_n_amps()):
                     overscan_level = _subtract_overscan_3d(image, i)
-                    logs.add_tag(logging_tags, 'OVERSCN{0}'.format(i + 1), overscan_level)
-                    self.logger.info('Subtracting overscan', extra=logging_tags)
+                    logger.info('Subtracting overscan', image=image,
+                                extra_tags={'OVERSCN{0}'.format(i + 1): float(overscan_level)})
```
cmccully: I think we can make this a single message at the end with all of the overscn keywords in one set. Right now 4 messages to the logger seems excessive.

banzai/logs.py
```diff
@@ -1,21 +1,45 @@
...
+class BanzaiLogger(logging.getLoggerClass()):
+    def __init__(self, name, level='NOTSET'):
+        super(BanzaiLogger, self).__init__(name, level)
+
+    def _log(self, level, msg, *args, **kwargs):
+        kwargs = _add_tag_dictionary(kwargs)
+        super(BanzaiLogger, self)._log(level, msg, *args, **kwargs)
+
+
+def _add_tag_dictionary(kwargs):
```
cmccully: maybe _add_tag_to_dictionary? You might also want to say this is logging tags to dictionary.

banzai/astrometry.py
turnerm: Done

banzai/bias.py
turnerm: Done

banzai/logs.py
turnerm: I agree that the name was not clear, but I have now called it `create_logging_tags_dictionary` because I think that describes what it's really doing

turnerm:
>  Do we still have to start the queue logger here? 

It seems not, so I've taken it out for now. But maybe it has some important purpose, let me know if that's the case...

> Is it possible to remove the start_logging function from main? How does the logging level get set?

Unfortunately I don't think so, because some initial logger configuration is needed, e.g. to set the level. However, at least by removing the queue logger there is no longer a need for the `stop_logging` function.

PR #137, opened 2018-10-08
Feature/calibs not required

turnerm:


cmccully:
Too stale at this point. We will rethink this feature and reimplement.

PR #139, opened 2018-10-14
Hotfix/table to fits

gmbrandt:
Previously, fits_to_hdu would not save Tables with columns which had lists (e.g. a list of coefficients) as an element. E.g. col1 = [[1,1,1],[1,1,1],...] would not save. This PR corrects that by properly inputting the format argument of fits.Column with the PI(max) argument which allows for variable column widths. 

I also fixed a bug in which regenerate_data_table_from_fits_hdu_list would load table.Column as an element of a list into the regenerated Table object. I fixed that so that it now recasts such objects as lists, as they should be.

### DIFF
--- a/banzai/utils/fits_utils.py
+++ b/banzai/utils/fits_utils.py
@@ -91,11 +91,7 @@ def table_to_fits(table):
     :param table: astropy table
     :return: fits BinTableHDU
     """
-    columns = [fits.Column(name=col.upper(), unit=str(table[col].unit),
-                           format=fits_formats(table[col]),
-                           array=table[col]) for col in table.colnames]
-    hdu = fits.BinTableHDU.from_columns(columns)
-
+    hdu = fits.BinTableHDU(table)
     # Put in the description keywords
     for k in hdu.header.keys():
         if 'TTYPE' in k:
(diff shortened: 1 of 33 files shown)


banzai/utils/fits_utils.py
```diff
@@ -91,11 +91,7 @@ def table_to_fits(table):
     :param table: astropy table
     :return: fits BinTableHDU
     """
-    columns = [fits.Column(name=col.upper(), unit=str(table[col].unit),
-                           format=fits_formats(table[col].dtype),
-                           array=table[col]) for col in table.colnames]
-    hdu = fits.BinTableHDU.from_columns(columns)
-
+    hdu = fits.BinTableHDU(table)
```
turnerm: This change will affect a couple of things in the photometry HDU of reduced data, which probably don't matter too much, but the main one to note is that the column names will no longer be capitalized in (e.g. `XPEAK` will now be `xpeak`). 

Also, for some reason count units were output as `ct`, but they are now `count`; and columns without units used to have `unit = 'None'` but now they don't have any unit attribute at all.

gmbrandt: Ah okay, thanks. Curtis is there some way to test whether these changes will break the pipeline upon deployment? Or are e2e and units the only pre-deploy test options?

cmccully: So the fits record array (which these BinTableHDU.data are) do not have case sensitive keys. I think in general the FITS convention is not supposed to be case sensitive. If it becomes an issue, we can change the column names to be uppercase.

PR #145, opened 2018-10-24
Hotfix/preview logging

turnerm:
The recent changes to the logging caused the preview pipeline to report a heartbeat once per second. This is fixed by manually setting the `amqp` logger level to INFO. 

The logger is now set up in `settings.py`, and the level is set by the root logger.

### DIFF
--- a/banzai/main.py
+++ b/banzai/main.py
@@ -369,7 +369,7 @@ def run_preview_pipeline():
 
     logs.set_log_level(args.log_level)
     # Need to keep the amqp logger level low, or else it send heartbeat check messages every second
-    logging.getLogger('amqp').setLevel('INFO')
+    logging.getLogger('amqp').setLevel(getattr(logging, 'INFO'))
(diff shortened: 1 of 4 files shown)


banzai/main.py
```diff
@@ -361,6 +366,11 @@ def run_preview_pipeline():
     args.preview_mode = True
     args.raw_path = None
     args.filename = None
+
+    logs.set_log_level(args.log_level)
+    # Need to keep the amqp logger level low, or else it send heartbeat check messages every second
+    logging.getLogger('amqp').setLevel('INFO')
```
cmccully: Shouldn't this be set to args.log_level?

turnerm: No, because if the `amqp` log level is `DEBUG`, then it will log a heartbeat check every second for each listener. 
Maybe instead I could set the log level to be max(INFO, args.log_level)?

PR #146, opened 2018-10-30
Feature/schedulable not required

turnerm:


banzai/main.py
```diff
@@ -131,6 +134,10 @@ def parse_args(selection_criteria, extra_console_arguments=None,
 
     logs.set_log_level(args.log_level)
 
+    if getattr(args, 'telescope_schedulable_not_required', False):
```
cmccully: why getattr here instead of just args.telescope_schedulable_not_required?

cmccully: P.S. That is a really long variable name.

banzai/main.py
```diff
@@ -131,6 +134,10 @@ def parse_args(selection_criteria, extra_console_arguments=None,
...
+        selection_criteria = [selection_criterion for selection_criterion in selection_criteria
+                              if not selection_criterion.__eq__(TelescopeCriterion('schedulable', operator.eq, True))]
+
```
cmccully: Should we add the schedulable criterion instead of deleting it? i.e. Not have it in the main and only add it in under reduce night later?

turnerm: My reasoning was that I wanted to keep the `IMAGING_CRITERIA` consistent with the default setting, since not requiring schedulability won't be the usual use case. But I'm happy to change it.

PR #149, opened 2018-11-05
Hotfix/schedulable reduce night

cmccully:
Fixes in how we run reduce_night.

### DIFF
--- a/banzai/main.py
+++ b/banzai/main.py
@@ -179,6 +179,16 @@ def process_directory(pipeline_context, raw_path, image_types=None, last_stage=N
                 logger.error(e, extra_tags={'filename': image})
 
 
+def process_single_frame(pipeline_context, raw_path, filename, last_stage=None, extra_stages=None, log_message=''):
+    if len(log_message) > 0:
+        logger.info(log_message, extra_tags={'raw_path': raw_path})
+    stages_to_do = get_stages_todo(last_stage, extra_stages=extra_stages)
+    try:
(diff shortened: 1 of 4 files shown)


banzai/main.py
```diff
@@ -259,17 +277,19 @@ def reduce_night():
 
     telescopes = dbs.get_telescopes_at_site(pipeline_context.site,
                                             db_address=pipeline_context.db_address,
-                                            must_be_schedulable=pipeline_context.telescopes_must_be_schedulable)
+                                            must_be_schedulable=not pipeline_context.ignore_schedulability)
```
turnerm: Oops, this was my bad! But maybe it would be less confusing if `must_be_schedulable` is changed to `ignore_schedulability` as well, i.e. to avoid passing a `not`.

cmccully: Ok. Done.

banzai/main.py
```diff
@@ -179,6 +179,16 @@ def process_directory(pipeline_context, raw_path, image_types=None, last_stage=N
                 logger.error(e, extra_tags={'filename': image})
 
 
+def process_single_frame(pipeline_context, raw_path, filename, last_stage=None, extra_stages=None, log_message=''):
+    if len(log_message) > 0:
+        logger.info(log_message, extra_tags={'raw_path': raw_path})
```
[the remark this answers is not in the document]
cmccully: Indeed. Added.

PR #150, opened 2018-11-06
Feature/refactor bpm

turnerm:
This PR mainly involves moving most BPM related methods from `image_utils.py` to the actual BPM stage.

In terms of functionality, the main difference is that the BPM is now read in and added to the image object only during the BPM stage, rather than as part of the image read-in process. However, the user should still experience the same behaviour through all of the entrypoint commands (the `--no-bpm` flag will have the same functionality as before), you will only find a difference if using the `read_images` command from `images.py` directly.


banzai/bpm.py
```diff
@@ -8,7 +20,72 @@ def group_by_keywords(self):
         return None
 
     def do_stage(self, images):
+        images_to_remove = []
         for image in images:
-            bpm_slices = array_utils.array_indices_to_slices(image.bpm)
-            image.bpm[image.data[bpm_slices] >= float(image.header['SATURATE'])] = 2
+            try:
+                self.add_bpm_to_image(image)
+            except MissingBPMError:
```
cmccully: I don't think you need to catch an exception here. I think this stage should just load a dummy bpm and set L1BPMID (or whatever the keyword is supposed to be) to an empty string. Then the code here can just check if that header keyword is an empty string and remove the image if necessary. I think that should simplify the logic quite a bit and make it easier in the long run to remove that entirely.

banzai/bpm.py
```diff
@@ -8,7 +20,72 @@ def group_by_keywords(self):
...
+            images.remove(image)
         return images
+
+    def add_bpm_to_image(self, image):
+        # Get the BPM filename
+        bpm_filename = self.get_bpm_filename(image)
+        # Check if file is missing
+        if bpm_filename is None:
+            logger.warning('Unable to find BPM in database', image=image)
+            raise MissingBPMError
```
cmccully: See above comment. I think this should load a dummy bpm, not raise an exception.

banzai/bpm.py
```diff
@@ -8,7 +20,72 @@ def group_by_keywords(self):
...
+        # Check if the BPM is the right size
+        if not bpm_has_valid_size(bpm, image):
+            logger.warning('BPM shape mismatch', image=image)
+            raise MissingBPMError
+        # Add BPM to image and header info
+        image.bpm = bpm
+        image.header['L1IDMASK'] = (os.path.basename(bpm_filename), 'Id. of mask file used')
+
+    def get_bpm_filename(self, image):
+        return dbs.get_bpm_filename(image.telescope.id, image.ccdsum,
```
cmccully: get_bpm_filename could return an empty string instead of None perhaps?

banzai/images.py
```diff
@@ -215,7 +215,7 @@ def read_images(image_list, pipeline_context):
                 raise dbs.TelescopeMissingException(error_message)
             munge(image, pipeline_context)
             if image.bpm is None:
-                image_utils.load_bpm(image, pipeline_context)
+                image_utils.add_empty_bpm(image)
```
cmccully: Perhaps just add image.bpm = None here? Or perhaps it is already None?

banzai/bpm.py
```diff
@@ -8,7 +16,80 @@ def group_by_keywords(self):
...
+        image.bpm = bpm
+        image.header['L1IDMASK'] = (os.path.basename(bpm_filename), 'Id. of mask file used')
+        logger.debug('Added BPM from file {}'.format(bpm_filename))
+
+    def get_bpm_filename(self, image):
+        return dbs.get_bpm_filename(image.telescope.id, image.ccdsum,
+                                    db_address=self.pipeline_context.db_address)
+
+
+def load_bpm(bpm_filename):
```
cmccully: Would it make sense to load the dummy bpm here instead of on image creation?

banzai/bpm.py
turnerm: I've taken out the exception, but I've changed the logic a bit from what you proposed: first, an attempt is made to add a BPM. If `image.bpm` is still `None` after this (meaning that the BPM could not be loaded or was the wrong size), then the stage checks whether `pipeline_context.no_bpm` is `True`, and if so adds an empty BPM; otherwise, the image is removed.

banzai/bpm.py
turnerm: The exception is now removed

banzai/bpm.py
turnerm: I've left this for now, did you have in mind that `add_bpm_to_image` would not short-circuit if the filename is None, but rather continue and try to load from an empty string? I couldn't get the logic for that to work in a nice way, but we can talk about it.

banzai/images.py
turnerm: It's already `None`, so I've now removed this line altogether.

PR #158, opened 2018-11-14
Feature/print full traceback

cmccully:
Added full traceback of uncaught exceptions to the logs.


cmccully:
My thought was to only print the full traceback when I caught a blanket exception. Basically if there was an exception that I knew I needed to handle I didn't bother with the full traceback because those are "expected" and therefore typically easy to find where the issue was. This was meant to mostly make it easier when an unexpected exception got raised.

PR #159, opened 2018-11-14
Fix/master bias level

turnerm:
The recent change to create a `CalibrationStacker` class resulted in one problem. The old `BiasMaker` class had a line to calculate the master bias level from the mean of all input images and add it to the master header as `BIASLVL`. This value is subtracted from any images that the master bias is applied to. With this step missing, an incorrect value for `BIASLVL` is written to the master bias header and used for subtraction. 

I've now added a `BIASLVL` computation to `make_master_calibration_frame` in the `BiasMaker` class. I also did the same for `FlatMasker`. This way all of the calibration makers require just a single stage, which works much better with the single image per stage and calibs from DB branches. I could be convinced to put `FlatMasker` back into its own stage, however, the `BIASLVL` computation cannot be a separate stage as it requires both the full image list and the final master image.

### DIFF
--- a/banzai/bias.py
+++ b/banzai/bias.py
@@ -27,12 +27,6 @@ class BiasMaker(CalibrationStacker):
     def min_images(self):
         return 5
 
-    def make_master_calibration_frmae(self, images):
-        master_image = super(BiasMaker, self).make_master_calibration_frame(images)
-        master_image.header['BIASLVL'] = (np.mean([image.header['BIASLVL'] for image in images]),
-                                          'Mean bias level of master bias')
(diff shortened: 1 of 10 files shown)


banzai/bias.py
```diff
@@ -25,6 +27,12 @@ def calibration_type(self):
     def min_images(self):
         return 5
 
+    def make_master_calibration_frame(self, images):
+        master_image = super(BiasMaker, self).make_master_calibration_frame(images)
```
cmccully: I like this much better.

PR #161, opened 2018-11-26
Hotfix/bpm mismatch and stage logging

turnerm:
This should fix a couple of issues that came up from all the recent changes: 
- In the BPM stage, I had not implemented the BPM filename header keyword check correctly, I've now changed it to use `header.get`
- If a stage was passed an empty list, it would crash on the logger which tried to get the 0th image
- `create_master_calibration_header` in `fits_utils.py` had an error in the logging call if a keyword could not be added

### DIFF
--- a/banzai/bpm.py
+++ b/banzai/bpm.py
@@ -17,7 +17,7 @@ class BPMUpdater(Stage):
         for image in images:
             add_bpm_to_image(image, self.pipeline_context)
             validate_bpm_size(image)
-            if image.header['L1IDMASK'][0] == '' and not self.pipeline_context.no_bpm:
+            if image.header.get('L1IDMASK') == '' and not self.pipeline_context.no_bpm:
                 logger.error("Can't add BPM to image, stopping reduction", image=image)
                 images_to_remove.append(image)
                 continue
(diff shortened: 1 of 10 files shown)


banzai/bpm.py
```diff
@@ -17,7 +17,7 @@ def do_stage(self, images):
         for image in images:
             add_bpm_to_image(image, self.pipeline_context)
             validate_bpm_size(image)
-            if image.header['L1IDMASK'][0] == '' and not self.pipeline_context.no_bpm:
+            if image.header.get('L1IDMASK') == '' and not self.pipeline_context.no_bpm:
```
cmccully: I would update to image.header.get('L1IDMASK', '')

banzai/stages.py
```diff
@@ -18,7 +18,8 @@ def stage_name(self):
         return '.'.join([__name__, self.__class__.__name__])
 
     def run(self, images):
-        logger.info('Running {0}'.format(self.stage_name), image=images[0])
+        if len(images):
```
cmccully: I know it is implied but I like len(images)  > 0

banzai/bpm.py
turnerm: Good idea, done

banzai/stages.py
turnerm: Fixed

PR #165, opened 2018-11-29
feature/refactor_settings

turnerm:
I've stripped all banzai-related imports from `settings.py` and made everything a string. Strings that represent classes or objects are parsed in `main.py`.

I've also restructured `settings.py` into a configuration file. There is a main `Settings` class that can have settings that are shared between imaging & NRES, e.g. the schedulability criterion. It also has a few fixed properties, e.g. selection criteria, that all subclasses must implement. The `Imaging` class then has imaging-specific settings. The version of settings you want can be specified from `main.py`.  

Now in `main.py`, `settings` is only referenced once, in the `parse_args` method. Here, a configuration object is created and passed directly to `PipelineContext`, where all attributes are added to the context object.

### DIFF
--- a/banzai/settings.py
+++ b/banzai/settings.py
@@ -1,7 +1,28 @@
 import operator
+from abc import ABC
 
 
-class Settings:
+class Settings(ABC):
+
+    @property
+    def SELECTION_CRITERIA(self):
+        raise NotImplementedError
+
+    @property
+    def IMAGE_CLASS(self):
+        raise NotImplementedError
+
+    @property
+    def ORDERED_STAGES(self):
+        raise NotImplementedError
+
+    @property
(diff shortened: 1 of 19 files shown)


banzai/settings.py
```diff
@@ -1,79 +1,102 @@
...
 
 from banzai.context import TelescopeCriterion
 from banzai import qc, bias, crosstalk, gain, mosaic, bpm, trim, dark, flats, photometry, astrometry, images
 
 
-# Logger set up
-logging.captureWarnings(True)
+class Settings(ABC):
+
+    @property
```
cmccully: I think you want to use the abstractmethod here instead of a not implemented error.

turnerm: Fixed

PR #168, opened 2018-12-06
Feature/calibs from db

turnerm:


banzai/calibrations.py
```diff
@@ -45,7 +45,7 @@ def run(self, images):
         return processed_images
 
     def do_stage(self, images):
-        min_images = self.pipeline_context.CALIBRATION_MIN_IMAGES.get(self.calibration_type.upper(), -1)
+        min_images = self.pipeline_context.CALIBRATION_MIN_FRAMES.get(self.calibration_type.upper(), -1)
```
cmccully: why is that -1 and not 0?

banzai/context.py
```diff
@@ -35,13 +35,13 @@ def __init__(self, command_line_args, settings,
             elif variable != 'command_line_args':
                 super(PipelineContext, self).__setattr__(variable, local_variables[variable])
 
-        for keyword in vars(command_line_args):
-            super(PipelineContext, self).__setattr__(keyword, getattr(command_line_args, keyword))
-
         for key, value in dict(inspect.getmembers(settings)).items():
             if not key.startswith('_'):
```
cmccully: I would add a comment in settings.py to say that anything prefaced with an underscore will get into pipeline context.

banzai/dbs.py
```diff
@@ -352,23 +352,22 @@ def get_bpm_filename(instrument_id, ccdsum, db_address=_DEFAULT_DB):
...
+                         'dateobs': image.dateobs,
+                         'datecreated': image.datecreated,
+                         'instrument_id': image.instrument.id,
+                         'is_master': image.is_master,
+                         'is_bad': image.is_bad,
                          'attributes': {}}
-    for image_attribute in image_attributes:
-        record_attributes['attributes'][image_attribute] = getattr(image_config, image_attribute)
+    for attribute in image.attributes:
+        record_attributes['attributes'][attribute] = str(getattr(image, attribute))
```
cmccully: I don't think we should cast to a string here.

banzai/dbs.py
```diff
@@ -426,21 +425,28 @@ def instrument_passes_criteria(instrument, criteria):
...
-    calibration_criteria &= CalibrationImage.dateobs < (image.dateobs + datetime.timedelta(days=365))
+        calibration_criteria &= cast(CalibrationImage.attributes[criterion], String) ==\
+                                type_coerce(getattr(image, criterion), JSON)
+
+    if realtime_reduction:
+        # We want to avoid using different master calibrations for the same block, therefore
+        # we make sure the the calibration frame used was created before the block start time
+        logger.debug("Realtime reduction mode is on, therefore we are only selecting master calibration"
+                     "frames with a creation date earlier than the block start date", image=image)
+        calibration_criteria &= CalibrationImage.datecrt < image.block_start
```
cmccully: datecrt.

banzai/dbs.py
```diff
@@ -426,21 +425,28 @@ def instrument_passes_criteria(instrument, criteria):
...
+    else:
+        # Only grab the last year. In principle we could go farther back, but this limits the number
+        # of files we get back. And if we are using calibrations that are more than a year old
+        # it is probably a bad idea anyway.
+        calibration_criteria &= CalibrationImage.dateobs < (image.dateobs + datetime.timedelta(days=365))
     calibration_criteria &= CalibrationImage.dateobs > (image.dateobs - datetime.timedelta(days=365))
```
cmccully: Maybe cut this, and add a warning (to grafana) when a calibration that is older than a month is used?

banzai/dbs.py
```diff
@@ -458,3 +464,56 @@ def get_master_calibration_image(image, calibration_type, master_selection_crite
...
     db_session.close()
     return calibration_file
+
+
+def get_individual_calibration_images(instrument, date_range, calibration_type, selection_criteria,
+                                      db_address=_DEFAULT_DB):
+
+    calibration_criteria = CalibrationImage.type == calibration_type.upper()
+    calibration_criteria &= CalibrationImage.instrument_id == instrument.id
+    calibration_criteria &= CalibrationImage.is_master.isnot(True)
```
cmccully: Let's add this as an option.

banzai/calibrations.py
turnerm: It was due to the possibility of there being 0 images, however I've refactored this section to give a more specific error in case the key is missing.

banzai/context.py
turnerm: Done

banzai/dbs.py
turnerm: Removed

banzai/dbs.py
turnerm: Fixed

banzai/dbs.py
turnerm: Do you mean to Kibana? DId you have in mind something like a QC parameter that tracks whether an old master has been used?

banzai/dbs.py
turnerm: Done

PR #169, opened 2018-12-21
Reafactor for BANZAI NRES

cmccully:
The master calibrations are now imported using the settings.FRAME_CLASS instead of being hard coded to images.Image. This required significant refactoring of the tests. The attributes in the db are now passed around as strings. Settings.py has been refactored to make it easier to override.


banzai/dbs.py
```diff
@@ -337,10 +337,9 @@ def get_instrument(header, db_address=_DEFAULT_DB):
 
 def get_bpm_filename(instrument_id, ccdsum, db_address=_DEFAULT_DB):
     db_session = get_session(db_address=db_address)
-    bpm_query = db_session.query(CalibrationImage).filter(
-                    CalibrationImage.type == 'BPM',
-                    CalibrationImage.instrument_id == instrument_id,
-                    cast(CalibrationImage.attributes['ccdsum'], String) == type_coerce(ccdsum, JSON))
+    bpm_query = db_session.query(CalibrationImage).filter(CalibrationImage.type == 'BPM',
+                                                          CalibrationImage.instrument_id == instrument_id,
+                                                          CalibrationImage.attributes['ccdsum'] == type_coerce(ccdsum, JSON))
```
turnerm: Are you sure you want to take remove `cast`? From [here](https://docs.sqlalchemy.org/en/latest/core/type_basics.html?highlight=json#sqlalchemy.types.JSON):
>Index operations return an expression object whose type defaults to JSON by default, so that further JSON-oriented instructions may be called upon the result type. Note that there are backend-specific idiosyncracies here, including that the PostgreSQL database does not generally compare a “json” to a “json” structure without type casts. These idiosyncracies can be accommodated in a backend-neutral way by making explicit use of the cast() and type_coerce() constructs. Comparison of specific index elements of a JSON object to other objects works best if the left hand side is CAST to a string and the right hand side is rendered as a JSON string

banzai/tests/utils.py
```diff
@@ -53,6 +53,7 @@ class FakeContext(object):
     def __init__(self, preview_mode=False, settings=banzai.settings.ImagingSettings()):
         self.processed_path = '/tmp'
         self.preview_mode = preview_mode
+        self.FRAME_CLASS = FakeImage
```
turnerm: I think `FRAME_CLASS` will be overwritten by the loop below

banzai/dbs.py
cmccully: Added back.

banzai/tests/bias_utils.py
```diff
@@ -0,0 +1,20 @@
+from banzai.tests.utils import FakeContext, FakeImage
+import numpy as np
+from astropy.io import fits
+
+
+class FakeBiasImage(FakeImage):
+    def __init__(self, context=None, data=None, bias_level=0.0, nx=101, ny=103, header=None):
+        super(FakeBiasImage, self).__init__(image_multiplier=bias_level, nx=nx, ny=ny, pipeline_context=context)
+        if data is not None:
```
turnerm: I think these if-statements can be removed since they are also executed in `FakeImage`. And maybe the default header can be set before `super` is called, and then passed directly. This comment also applies to `FakeDarkImage` and `FakeFlatImage`.

cmccully: Done

banzai/tests/bias_utils.py
```diff
@@ -0,0 +1,23 @@
...
+        super(FakeBiasImage, self).__init__(image_multiplier=bias_level, nx=nx, ny=ny)
+        if data is not None:
+            self.data = data
+        if header is None:
+            self.header = fits.Header({'BIASLVL': bias_level, 'OBSTYPE': 'BIAS', 'TELESCOP': '1m0-10'})
+        else:
+            self.header = header
+
+
+def make_context_with_master_bias(bias_level, readnoise, nx, ny):
```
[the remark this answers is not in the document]
cmccully: Done.

PR #173, opened 2019-01-23
Feature/calimages datecrt field

turnerm:


banzai/utils/fits_utils.py
```diff
@@ -45,6 +46,7 @@ def create_master_calibration_header(images):
     mean_dateobs = date_utils.mean_date(observation_dates)
 
     header['DATE-OBS'] = (date_utils.date_obs_to_string(mean_dateobs), '[UTC] Mean observation start time')
+    header['DATE-CRT'] = (date_utils.date_obs_to_string(datetime.datetime.utcnow()), '[UTC] Master creation time')
```
cmccully: Often, the convention is that DATE is the date the fits file was created.

banzai/dbs.py
```diff
@@ -67,6 +67,7 @@ class CalibrationImage(Base):
     filename = Column(String(50), unique=True)
     filepath = Column(String(100))
     dateobs = Column(DateTime, index=True)
+    datecrt = Column(DateTime, index=True)
```
cmccully: I dislike datecrt. Why not datecreated?

banzai/images.py
```diff
@@ -166,6 +168,9 @@ def write_catalog(self, filename, nsources=None):
     def add_history(self, msg):
         self.header.add_history(msg)
 
+    def upate_datecrt(self):
```
cmccully: Why does this need its own method?

banzai/utils/fits_utils.py
turnerm: I switched to using `DATE`

banzai/dbs.py
turnerm: Changed to `datecreated`

banzai/images.py
turnerm: I wanted to avoid directly modifying an object attribute.

PR #176, opened 2019-02-01
Hotfix/block date na

cmccully:
This fixes an issue when the block date is set to N/A because the image was taken by the control interface instead of being taken as part of a normal block.

### DIFF
--- a/banzai/utils/date_utils.py
+++ b/banzai/utils/date_utils.py
@@ -82,6 +82,8 @@ def parse_epoch_string(epoch_string):
 
 
 def parse_date_obs(date_obs_string):
+    if date_obs_string == 'N/A':
+        return None
     # Check if there are hours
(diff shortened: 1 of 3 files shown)


banzai/utils/date_utils.py
```diff
@@ -82,6 +82,8 @@ def parse_epoch_string(epoch_string):
 
 
 def parse_date_obs(date_obs_string):
+    if date_obs_string == 'N/A':
+        return None
```
turnerm: Maybe log a warning in this case?

cmccully: I have mixed feelings on this. I think not because all frames that are taken manually will produce a warning. I think if there is a problem, the exception will be self explanatory. If we don't need it, we don't need to clutter the logs.

PR #182, opened 2019-02-08
Feature/remove preview

cmccully:
This branch's goal is to separate the preview mode from the concept of realtime processing. I have done some refactoring to ensure that we don't try to process frames that we can't like guider frames. I have also done some refactoring to make things more testable.


banzai/realtime.py
```diff
@@ -0,0 +1,75 @@
...
+    context: iterable
+             A pipeline context with FRAME_SELECTION_CRITERIA
+    db_address: str
+                SQLAlchemy style URL to the database with the status of previous reductions
+    max_tries: int
+               Maximum number of retries to reduce an image
+
+    Returns
+    -------
+    need_preview: bool
```
turnerm: Should be `need_to_process`

banzai/main.py
```diff
@@ -155,7 +155,7 @@ def process_directory(pipeline_context, raw_path, image_types=None, log_message=
     if len(log_message) > 0:
         logger.info(log_message, extra_tags={'raw_path': raw_path})
     image_path_list = image_utils.make_image_path_list(raw_path)
-    image_path_list = image_utils.select_images(image_path_list, pipeline_context.FRAME_SELECTION_CRITERIA,
+    image_path_list = image_utils.select_images(image_path_list, pipeline_context,
                                                 image_types=image_types, db_address=pipeline_context.db_address)
```
turnerm: The keyword arguments `image_types` and `db_address` should be removed

banzai/context.py
```diff
@@ -1,5 +1,5 @@
 import inspect
-
+from banzai.utils import image_utils
```
turnerm: Missing a blank line

banzai/context.py
```diff
@@ -47,3 +47,6 @@ def __delattr__(self, item):
 
     def __setattr__(self, key, value):
         raise TypeError('Resetting attribute is not allowed. PipelineContext is immutable.')
+
```
turnerm: On line 24, change `preview_mode` keyword to `realtime_reduction`

banzai/utils/image_utils.py
```diff
@@ -2,42 +2,28 @@
...
-    obstype = None
-    hdu_list = fits.open(filename)
-    for hdu in hdu_list:
-        if 'OBSTYPE' in hdu.header.keys():
-            obstype = hdu.header['OBSTYPE']
-    if obstype is None:
-        logger.error('Unable to get OBSTYPE', extra_tags={'filename': filename})
-    return obstype
+def image_can_be_processed(header, context):
+    passes = dbs.instrument_passes_criteria(header, context.FRAME_SELECTION_CRITERIA)
```
turnerm: `dbs.instrument_passes_criteria` needs to be updated to handle a header, rather than an instrument object

banzai/main.py
```diff
@@ -297,20 +297,23 @@ def run_end_of_night():
             logger.error(logs.format_exception())
```
turnerm: `run_end_of_night` should be removed, as it is not required once we start realtime reductions, and furthermore it is now broken by the removal of `image_type` from `image_utils.select_images`.

banzai/context.py
cmccully: Ok. I guess so. I think this whole structure merits a little thought in a future PR.

banzai/realtime.py
cmccully: Done

banzai/main.py
cmccully: Done.

banzai/context.py
cmccully: Done. I set the default to False. This also makes me think we should rethink this whole pattern.

banzai/utils/image_utils.py
cmccully: Done.

turnerm: Also, if `instrument_passes_criteria` no longer queries the DB, then it should probably be moved out of `dbs.py`

cmccully:
I think I fixed it.

banzai/main.py
```diff
@@ -155,9 +155,11 @@ def process_directory(pipeline_context, raw_path, image_types=None, log_message=
     if len(log_message) > 0:
         logger.info(log_message, extra_tags={'raw_path': raw_path})
     image_path_list = image_utils.make_image_path_list(raw_path)
-    image_path_list = image_utils.select_images(image_path_list, pipeline_context,
-                                                image_types=image_types, db_address=pipeline_context.db_address)
-    for image_path in image_path_list:
+    if image_types is None:
+        image_types = [None]
+    images_to_reduce = [image_utils.select_images(image_path_list, pipeline_context, image_type)
```
turnerm: `image_utils.select_images()` needs to be updated to handle the `image_type` parameter

PR #183, opened 2019-02-11
fix/instrument fully specified

turnerm:
I realized that even though we added the per-instrument tracking of `enclosure` and `telescope`, those parameters are not queried when an instrument is searched for. This fixes the problem somewhat, although currently `main.stack_calibrations()` only takes `instrument` (as in `camera`) and `site` as parameters. 

Another potential issue: BPMs are currently stored in directories structured with as `/site/camera/`. Thus, if a camera moves enclosures / telescopes within the same site, BANZAI will require a new BPM for reduction, and this will need to be updated manually in the database (but the BPM file itself will be the same).


banzai/dbs.py
```diff
@@ -82,11 +82,12 @@ class Instrument(Base):
     __tablename__ = 'instruments'
     id = Column(Integer, primary_key=True, autoincrement=True)
     site = Column(String(15), ForeignKey('sites.id'), index=True)
```
dmcollom: Shouldn't site also be nullable=False, or is that done automatically by virtue of being declared a foreign key?

banzai/dbs.py
```diff
@@ -296,43 +307,38 @@ class SiteMissingException(Exception):
...
+    if must_be_schedulable:
+        criteria &= Instrument.schedulable.is_(True)
+    instrument = db_session.query(Instrument).filter(criteria).order_by(Instrument.id.desc()).first()
     db_session.close()
-    instrument = query_for_instrument(db_address, site, camera)
     return instrument
 
 
-def get_instrument(header, db_address=_DEFAULT_DB):
+def get_instrument(header, db_address=_DEFAULT_DB, configdb_address='http://configdb.lco.gtn/sites/'):
```
dmcollom: Better to put the configdb_address in a constant up top, I think.

PR #188, opened 2019-02-21
Fix/only open fits files

turnerm:


CHANGES.md
```diff
@@ -1,3 +1,7 @@
+0.19.3 (2019-02-14)
```
cmccully: should be 0.19.4

PR #189, opened 2019-02-22
Schedule stacking from lake

cmccully:
As part of the migration to realtime processing, we want to stack the new calibrations as soon as possible after completion. To this end, we query the lake to figure out when the next bias/dark/flat is done so we know when to make new stacks. As the scheduling is getting more complicated, we are also setting up dramatiq here.


banzai/main.py
```diff
@@ -419,3 +420,66 @@ def update_db():
...
+                                                                      use_masters=False,
+                                                                      db_address=runtime_context.db_address))
+    expected_image_count = 0
+    for block in blocks:
+        for molecule in block['molecules']:
+            if calibration_type.upper() == molecule['type']:
+                expected_image_count += molecule['exposure_count']
+    logger.info('expected image count: {0}'.format(str(expected_image_count)))
+    if (expected_image_count < completed_image_count and not process_any_images):
+        raise
```
turnerm: There's no active exception here to raise

banzai/main.py
```diff
@@ -419,3 +420,66 @@ def update_db():
...
+    else:
+        process_master_maker(runtime_context, instrument, calibration_type, runtime_context.min_date,
```
turnerm: Shouldn't this be `process_master_maker.send(...`

setup.cfg
```diff
@@ -83,7 +84,7 @@ version = 0.20.1
 banzai = banzai.main:main
 banzai_reduce_individual_frame = banzai.main:reduce_single_frame
 banzai_reduce_directory = banzai.main:reduce_directory
-banzai_stack_calibrations = banzai.main:stack_calibrations
```
turnerm: Do we maybe want to retain a method for stacking calibrations locally, i.e. an entrypoint that just calls `process_master_maker(...)` without the `.send` (I think this would work...)

dmcollom: You're completely right--I've added a wrapper function with an entrypoint that calls process_master_maker.

banzai/main.py
dmcollom: process_master_maker is already being called from an asynchronous worker process, so used .send() would spawn off yet another thread. Probably best not to do that.

banzai/main.py
dmcollom: You are correct. Updated.

banzai/images.py
```diff
@@ -194,10 +194,10 @@ def _add_bpm_to_hdu_list(self, hdu_list):
             hdu_list.append(bpm_hdu)
         return hdu_list
 
-    def _post_to_archive(self, filepath):
+    def _post_to_archive(self, filepath, pipeline_context):
```
turnerm: This should be updated to `runtime_context`

banzai/images.py
```diff
@@ -194,10 +194,10 @@ def _add_bpm_to_hdu_list(self, hdu_list):
...
         logger.info('Posting file to the archive', image=self)
         try:
-            file_utils.post_to_archive_queue(filepath)
+            file_utils.post_to_archive_queue(filepath, pipeline_context.broker_url)
```
turnerm: Also here, `pipeline_context` should be renamed `runtime_context`

banzai/utils/date_utils.py
```diff
@@ -135,16 +135,14 @@ def valid_date(s):
 
 def get_min_and_max_dates(timezone, dayobs, return_string=False):
     midnight_at_site = _get_midnight(timezone, dayobs)
-    min_date = midnight_at_site - datetime.timedelta(days=0.5)
-    max_date = midnight_at_site + datetime.timedelta(days=0.5)
+    min_date = midnight_at_site + datetime.timedelta(days=0.5)
+    max_date = midnight_at_site + datetime.timedelta(days=1.5)
```
turnerm: I think this needs to be set to `min_date = midnight_at_site - timedelta(days=1)` and `max_date = midnight_at_site`. 

The reason this method was written to span midnight by +/- 0.5 days is because it has so far been used only in the e2e tests. The data for these tests is taken from when the same skyflat filter was used for evening / morning flats on the same night. Thus, for the tests you need to get flats from 1/2 a day before and after midnight. 

These days, you want to get the flats from the previous morning (so midnight minus one day) and then this evening (so up until this midnight.

banzai/utils/date_utils.py
```diff
@@ -135,16 +135,14 @@ def valid_date(s):
...
     return min_date, max_date
 
 
+# Returns previous midnight relative to date of observation
 def _get_midnight(timezone, dayobs):
-    midnight_in_utc = datetime.datetime.combine(epoch_string_to_date(dayobs) + datetime.timedelta(days=1),
-                                                datetime.time(0, 0, 0))
-    local_midnight = midnight_in_utc - datetime.timedelta(hours=timezone)
-    return local_midnight
+    return datetime.datetime.strptime(dayobs, '%Y%m%d') - datetime.timedelta(hours=timezone)
```
turnerm: Doesn't it make more sense to get "this" midnight, as in the one associated with the current dayobs?

At any rate, this change affects my above comment about `get_min_and_max_dates`. In this case they should be:
`min_date = midnight_at_site`
`max_date = midnight_at_site + timedelta(days=1)`

banzai/main.py
```diff
@@ -91,7 +100,7 @@ def parse_args(extra_console_arguments=None, parser_description='Process LCO dat
                         default=False)
     parser.add_argument('--post-to-elasticsearch', dest='post_to_elasticsearch', action='store_true',
                         default=False)
-    parser.add_argument('--fpack', dest='fpack', action='store_true', default=False,
+    parser.add_argument('--fpack', dest='fpack', action='store_true', default=True,
```
turnerm: I think if you should remove `action='store_true'` or else it will not be possible to set `fpack` to False

banzai/celery.py
```diff
@@ -0,0 +1,113 @@
...
+    if (completed_image_count < expected_image_count and self.request.retries < 3):
+        raise self.retry()
+    else:
+        calibrations.process_master_maker(runtime_context, instrument, runtime_context.frame_type,
+                                          runtime_context.min_date, runtime_context.max_date)
+
+
+@app.task(name='celery.process_image')
+def process_image(path, runtime_context_dict):
+    logger.info('Got into actor.')
```
mgdaily: Looks like an extraneous log message?

dmcollom:
This has been closed due to the opening of a new branch in order to more practically squash commits.

PR #190, opened 2019-02-26
Feature/refactor context

cmccully:
We refactored the pipeline context into two pieces. There is the code level "settings" parameters and then there are the runtime parameters set by the users. This is so we can serialize all of the runtime parameters to post on a redis queue.


banzai/tests/test_need_to_process_image.py
```diff
@@ -22,7 +17,11 @@ def __init__(self, success=False, checksum=md5_hash1, tries=0):
...
 @mock.patch('banzai.realtime.get_primary_header')
-def test_no_processing_if_previous_success(mock_header, mock_processed, mock_md5):
+@mock.patch('banzai.dbs.get_instrument')
+@mock.patch('banzai.utils.image_utils.image_can_be_processed')
+def test_no_processing_if_previous_success(mock_can_process, mock_instrument, mock_header, mock_processed, mock_md5):
+    mock_can_process.return_value = True
+    mock_instrument.return_value = FakeInstrument()
     mock_processed.return_value = FakeRealtimeImage(success=True, checksum=md5_hash1)
     mock_md5.return_value = md5_hash1
     assert not need_to_process_image('test.fits', FakeContext())
```
turnerm: I think some of these tests need to be update to pass `ignore_schedulability` instead of a `PipelineContext` instance to `need_to_process_image`. Should we be worried that they're passing anyway?

banzai/realtime.py
```diff
@@ -21,16 +21,16 @@ def increment_try_number(path, db_address=dbs._DEFAULT_DB):
     dbs.commit_processed_image(image, db_address=db_address)
 
 
-def need_to_process_image(path, context, db_address=dbs._DEFAULT_DB, max_tries=5):
+def need_to_process_image(path, ignore_schedulability, db_address=dbs._DEFAULT_DB, max_tries=5):
```
turnerm: Would it make sense to set `ignore_schedulability` to `False` by default?

banzai/main.py
```diff
@@ -117,125 +121,119 @@ def parse_args(settings, extra_console_arguments=None,
...
-    return pipeline_context, raw_path
+            raw_path = runtime_context.raw_path
+    return runtime_context, raw_path
 
 
-def reduce_directory(pipeline_context=None, raw_path=None, image_types=None):
+def reduce_directory(runtime_context=None, raw_path=None, image_types=None):
     # TODO: Remove image_types once reduce_night is not needed
-    pipeline_context, raw_path = parse_directory_args(pipeline_context, raw_path, banzai.settings.ImagingSettings())
+    pipeline_context, raw_path = parse_directory_args(runtime_context, raw_path)
```
turnerm: Maybe rename this to `runtime_context` for consistency?

banzai/main.py
```diff
@@ -361,9 +355,9 @@ def run_individual_listener(broker_url, queue_name, pipeline_context):
 
 
 class RealtimeModeListener(ConsumerMixin):
-    def __init__(self, broker_url, pipeline_context):
+    def __init__(self, broker_url, runtime_context):
         self.broker_url = broker_url
-        self.pipeline_context = pipeline_context
+        self.pipeline_context = runtime_context
```
turnerm: I think this should be `self.runtime_context`

banzai/dbs.py
```diff
@@ -411,7 +411,7 @@ def get_instruments_at_site(site, db_address=_DEFAULT_DB, ignore_schedulability=
 
 
 def get_master_calibration_image(image, calibration_type, master_selection_criteria,
-                                 realtime_reduction=False, db_address=_DEFAULT_DB):
+                                 use_older_calibrations=False, db_address=_DEFAULT_DB):
```
turnerm: I found the `use_older_calibrations` variable name a bit ambiguous. What about something like `use_only_past_calibrations` or something like that?

banzai/settings.py
```diff
@@ -7,122 +7,77 @@
...
+            cal_file += '-{}'.format(filename_function(image))
+        cal_file += '.fits'
+        return cal_file
+    return get_calibration_filename
+
+
+CALIBRATION_FILENAME_FUNCTIONS = {'BIAS': make_calibration_filename_function('BIAS', [ccdsum_to_filename], telescope_to_filename),
+                                  'DARK': make_calibration_filename_function('DARK', [ccdsum_to_filename], telescope_to_filename),
+                                  'SKYFLAT': make_calibration_filename_function('SKYFLAT', [ccdsum_to_filename, filter_to_filename],
+                                                                                telescope_to_filename)}
```
turnerm: It seems that the two big changes to `settings.py` are that 1) the required settings parameters are no longer "enforced" since they are no longer properties, 2) there is no longer a distinction between general vs. imaging settings. 

Given point 2, how will the settings be implemented in BANZAI NRES? Have you thought about adding some kind of `settings_type` variable to `realtime_context`, which would specify the settings import type? Then the two sets of settings could remain distinct.

banzai/utils/image_utils.py
```diff
@@ -18,12 +27,17 @@ def get_reduction_level(header):
...
     images = []
     for filename in image_list:
         try:
             header = get_primary_header(filename)
-            if context.image_can_be_processed(header) and (image_type is None or get_obstype(header) == image_type):
+            should_process = image_can_be_processed(header, db_address)
+            should_process &= (image_type is None or get_obstype(header) == image_type)
+            if not ignore_schedulability:
+                instrument = dbs.get_instrument(header, db_address=db_address)
+                should_process &= instrument.schedulable
```
turnerm: I think there is some overlap with some of the things checked `realtime.need_to_process`, maybe some methods could be consolidated?

banzai/main.py
```diff
@@ -117,125 +121,119 @@ def parse_args(settings, extra_console_arguments=None,
...
+def process_directory(runtime_context, raw_path, image_types=None, log_message=''):
     if len(log_message) > 0:
         logger.info(log_message, extra_tags={'raw_path': raw_path})
     image_path_list = image_utils.make_image_path_list(raw_path)
     if image_types is None:
         image_types = [None]
     images_to_reduce = []
     for image_type in image_types:
-        images_to_reduce += image_utils.select_images(image_path_list, pipeline_context, image_type)
+        images_to_reduce += image_utils.select_images(image_path_list, runtime_context.db_address, image_type)
```
turnerm: The arguments for `select_images` defined in `image_utils` are different than the ones you passed here

banzai/settings.py
turnerm: After discussion I realized that this is not really possible so don't worry about this comment.

banzai/astrometry.py
```diff
@@ -32,6 +32,49 @@ def do_stage(self, image):
...
+
+                # Update the RA and Dec header keywords
+                image.header['RA'], image.header['DEC'] = get_ra_dec_in_sexagesimal(image.header['CRVAL1'],
+                                                                                    image.header['CRVAL2'])
+
+                # Clean up wcs file
+                os.remove(wcs_name)
+
+                # Add the RA and Dec values to the catalog
+                add_ra_dec_to_catalog(image)
```
turnerm: Was the merge with the GAIA changes done correctly? I wouldn't expect any major updates to `astrometry.py`

banzai/tests/test_pipeline_context.py
```diff
@@ -1,19 +1,11 @@
 from argparse import Namespace
```
turnerm: Could this test be renamed to `test_runtime_context.py` ?

PR #193, opened 2019-03-04
Feature/easy reprocess

turnerm:
After making the switch to reducing frames individually, it has been a bit cumbersome to reprocess a missed directory of data. In particular, the calibration have to be reduced individually, then stacked, and then finally the standard / expose frames can be reduced. This PR adds an entrypoint that goes through the above steps. To simplify its usage as much as possible, it is set to use all the default parameters used in production -- the only thing the user needs to provide is a directory and a database address.


banzai/main.py
```diff
@@ -222,9 +222,12 @@ def parse_directory_args(runtime_context=None, raw_path=None, extra_console_argu
 
 def reduce_directory(runtime_context=None, raw_path=None, image_types=None):
     # TODO: Remove image_types once reduce_night is not needed
+    pipeline_context, raw_path = parse_directory_args(runtime_context, raw_path)
```
cmccully: Why pipeline context and runtime context?

banzai/utils/date_utils.py
```diff
@@ -133,10 +133,14 @@ def valid_date(s):
         raise argparse.ArgumentTypeError(msg)
 
 
-def get_min_and_max_dates(timezone, dayobs, return_string=False):
+def get_min_and_max_dates(timezone, dayobs, return_string=False, single_night=False):
+    if single_night:
+        days_minus, days_plus = 0.5, 0.5
```
cmccully: This feels more confusing than it needs to be. Why not just do the min_date and max_date calculations inside the if statement?

banzai/main.py
turnerm: Oops, this line was a remnant from the merge, I've removed it

banzai/utils/date_utils.py
turnerm: Good point, I've changed it!

cmccully:
Superseded by changes to banzai-web

PR #194, opened 2019-03-05
Don't override ok SATURATE values

cmccully:
Updated munge code to only take the default values for saturation if there is no real value in the header.


banzai/munge.py
```diff
@@ -19,18 +19,21 @@ def munge(image):
 
     # 1m SBIGS
     elif '1m0' in image.instrument.type:
-        image.header['SATURATE'] = (46000.0, '[ADU] Saturation level used')
+        if image.header.get('SATURATE', 0.0) == 0.0:
+            image.header['SATURATE'] = (46000.0, '[ADU] Saturation level used')
     elif '0m4' in image.instrument.type or '0m8' in image.instrument.type:
-        image.header['SATURATE'] = (56000.0, '[ADU] Saturation level used')
```
drhaz: For consistency, do the  same treatment for 'MAXLIN' keyword with the identical default value of 64000 if not present.

PR #195, opened 2019-03-07
Feature/calibs not required 2

turnerm:
The `--no-bpm` flag for entrypoints is now changed to `--calibrations-not-required`, to include all calibration types: BPM, bias, dark and flat. 

Previously, if a master calibration was missing, the pipeline would continue to the next stage but mark the frame as bad. I'm not sure if that's the behaviour we wanted -- this is fine for individual calibration frames, but for science frames getting marked as bad is currently inconsequential (i.e., the science frame will still get uploaded to the archive without any obvious signal to the user that it was not properly reduced). 

In this PR, any missing calibration will stop the reduction, unless the `--calibrations-not-required` flag is set.

Also, there were no tests for applying darks and flats to science frames, so I've added those in.


cmccully:
Superseded by #231

PR #196, opened 2019-03-11
Feature/reduce stacking memory

turnerm:
This PR breaks up the reading of image data into subsets of rows so that the memory usage of stacking operations can be limited. The current approximate memory upper bound is set by the` nbytes_max` parameter in `CalibrationMaker.stack_calibrations()`.


cmccully:
Fixed in #231

PR #198, opened 2019-03-13
hotfix/dont stop due to inhomogeneity

turnerm:
This PR takes care of the case where the dimensions of an instrument are changed. When this happens, BANZAI will try to use a previous master calibration, but this will fail when the dimensions are found to be different.

A more permanent solution is found in PR #195.


cmccully:
Superseded by #231

PR #200, opened 2019-03-17
Fix/dbs for nres

gmbrandt:
A temporary fix for how banzai pulls nres data from LCO's configdb, which allows NRES data to reduce. The problems are: NRES frames have no keyword 'ENCID' in their primary header, 'TELID' is set to 'igla' (what the enclosure id should be). This fix makes it so that the database query does not need telescope or enclosure if it is an NRES instrument.


banzai/dbs.py
```diff
@@ -309,13 +329,17 @@ class SiteMissingException(Exception):
     pass
 
 
-def query_for_instrument(db_address, site, camera, enclosure, telescope, must_be_schedulable=False):
+def query_for_instrument(db_address, site, camera, enclosure=None, telescope=None,
```
turnerm: I would be hesitant to have `enclosure` and `telescope` default to `None`, since its fairly important for BANZAI imaging that they are specified, or you may end up returning the wrong instrument.

What if instead you relax the constraint on `enclosure` and `telescope` if `camera` contains the string `"nres"`?

banzai/dbs.py
```diff
@@ -149,22 +150,59 @@ def parse_configdb(configdb_address=_CONFIGDB_ADDRESS):
...
+                                      'schedulable': ins['state'] == 'SCHEDULABLE'}
+                        # hotfix for configdb
+                        if instrument['name'] is None:
+                            instrument['name'] = instrument['camera']
+                        if instrument['name'] in CAMERAS_FOR_INSTRUMENTS:
+                            instrument['camera'] = CAMERAS_FOR_INSTRUMENTS[instrument['name']]
+                        instruments.append(instrument)
+                        if 'nres' in instrument['name']:
+                            # hotfix so that we can reduce old nres frames when the cameras were flXX instead of faXX
+                            fl_instrument = instrument.copy()
```
cmccully: Deal with this in mock configdb data rather than here.

banzai/dbs.py
```diff
@@ -157,22 +158,54 @@ def parse_configdb(configdb_address=_CONFIGDB_ADDRESS):
...
+                            instrument['name'] = instrument['camera']
+                        if instrument['name'] in CAMERAS_FOR_INSTRUMENTS:
+                            instrument['camera'] = CAMERAS_FOR_INSTRUMENTS[instrument['name']]
+                        instruments.append(instrument)
+
+    instruments = remove_nres_duplicates(instruments)
     return sites, instruments
 
 
+def remove_nres_duplicates(instruments):
```
cmccully: It feels like this can be simplified.

PR #204, opened 2019-04-17
Fix/banzai memory issues

dmcollom:


banzai/celery.py
```diff
@@ -1,24 +1,23 @@
...
-from banzai.context import Context, ContextJSONEncoder
-from banzai.exceptions import IncompleteProcessingException
+from banzai.context import Context
 
-
-dramatiq.set_broker(settings.REDIS_BROKER)
-dramatiq.set_encoder(ContextJSONEncoder())
+app = Celery('banzai')
+app.config_from_object('banzai.celeryconfig')
+app.conf.update(broker_url=os.getenv('REDIS_HOST', 'redis://redis:6379/0'))
```
cmccully: should this be redis://localhost:6379/0 ?

banzai/celeryconfig.py
```diff
@@ -0,0 +1,4 @@
+broker_url = 'redis://redis:6379/0'
```
cmccully: here too?

banzai/context.py
```diff
@@ -13,6 +13,6 @@ def Context(args):
     return constructor(**args_dict)
 
 
-class ContextJSONEncoder(JSONEncoder):
-    def encode(self, data):
-        return json.dumps(data, separators=(",", ":"), default=str).encode("utf-8")
+# class ContextJSONEncoder(JSONEncoder):
+#     def encode(self, data):
```
cmccully: I think we are finally safe to remove these?

PR #208, opened 2019-05-09
Feature/readme rtd

cmccully:


cmccully:
See here for built version of read the docs.

https://banzai.readthedocs.io/en/feature-readme_rtd/

PR #216, opened 2019-07-08
using different library to parse datetimes

dmcollom:


Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG MINICONDA_VERSION
-FROM docker.lco.global/docker-miniconda3:${MINICONDA_VERSION}
```
cmccully: Was this supposed to get committed?

dmcollom: It was not, but nice rhetorical question. I've updated it.

PR #218, opened 2019-07-10
Add WCSERR keyword to images missing catalogs.

mgdaily:
This is a fix for the issue detailed in https://lcoglobal.redmineup.com/issues/966

When the astrometry stage encounters an image without a catalog of sources, the `WCSERR` keyword should be correctly populated (set to 4). All images detailed in the issue were missing `WCSERR` because the photometry stage had failed to generate a catalog.

### DIFF
--- a/banzai/astrometry.py
+++ b/banzai/astrometry.py
@@ -12,6 +12,9 @@ from banzai import settings
 
 logger = logging.getLogger('banzai')
 
+FAILED_WCS = (4, 'Error status of WCS fit. 0 for no error')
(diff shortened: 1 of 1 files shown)


banzai/astrometry.py
```diff
@@ -30,6 +30,7 @@ def do_stage(self, image):
 
         # Short circuit
         if image_catalog is None:
+            image.header['WCSERR'] = (4, 'Error status of WCS fit. 0 for no error')
```
dmcollom: This error message is used in 5+ places. It should either go in a constant at the top of the file, or we should take the opportunity to create an errors.py to centralize error messages (also a great way to standardize them).

PR #219, opened 2019-07-11
updated e2e tests to use mocked lake endpoint

dmcollom:


banzai/tests/test_end_to_end.py
```diff
@@ -70,16 +74,17 @@ def run_stack_calibrations(frame_type):
         raw_path = os.path.join(DATA_ROOT, day_obs, 'raw')
```
cmccully: This is no longer needed. Also, could you use DATA_ROOT below when you define raw_path rather than defining it it here?

dmcollom: Fixed.

PR #220, opened 2019-07-15
Add dbs.get_site function

Ara-yjx:
Add dbs.get_site function
param: db_address (default: _DEFAULT_DB)
return: list of distinct site codes in the database

Unit test included.

Collab with @xuiris

### DIFF
--- a/banzai/dbs.py
+++ b/banzai/dbs.py
@@ -435,7 +435,7 @@ def get_sites(db_address=_DEFAULT_DB):
     with get_session(db_address=db_address) as db_session:
         sites = db_session.query(Instrument.site).distinct()
     return [site[0] for site in sites]
-        
+
 
 def get_instruments_at_site(site, db_address=_DEFAULT_DB, ignore_schedulability=False):
     with get_session(db_address=db_address) as db_session:
--- a/banzai/dbs.py
+++ b/banzai/dbs.py
@@ -433,7 +433,7 @@ def get_timezone(site, db_address=_DEFAULT_DB):
 
 def get_sites(db_address=_DEFAULT_DB):
     with get_session(db_address=db_address) as db_session:
-        sites = db_session.query(Instrument.site).distinct()
+        sites = db_session.query(Site.id)
     return [site[0] for site in sites]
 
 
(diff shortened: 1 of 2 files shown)


banzai/dbs.py
```diff
@@ -431,6 +431,12 @@ def get_timezone(site, db_address=_DEFAULT_DB):
     return site_list[0].timezone
 
 
+def get_sites(db_address=_DEFAULT_DB):
+    with get_session(db_address=db_address) as db_session:
+        sites = db_session.query(Instrument.site).distinct()
```
cmccully: Why are you not just querying the sites table?

banzai/tests/test_dbs.py
```diff
@@ -84,3 +85,10 @@ def test_not_removing_singlets():
     assert len(culled_list) == 2
     assert culled_list[0]['name'] == 'nres01'
     assert culled_list[1]['name'] == 'cam01'
+
+def test_get_sites():
+    result = get_sites(db_address='sqlite:///test.db')
+    correct = ['bpl', 'coj', 'cpt', 'elp', 'lsc', 'ogg', 'sqa', 'tfn', 'tst']
```
cmccully: You should mock the call to the configdb here.

banzai/dbs.py
```diff
@@ -431,6 +431,12 @@ def get_timezone(site, db_address=_DEFAULT_DB):
     return site_list[0].timezone
 
 
+def get_sites(db_address=_DEFAULT_DB):
+    with get_session(db_address=db_address) as db_session:
+        sites = db_session.query(Site.id)
+    return [site[0] for site in sites]
```
cmccully: This should db_session.query(Site.id).all() rather than a list comprehension. Is this needed in banzai or is something this straightforward better in the banzai-web app?

xuiris: I have changed it to all(). I thought it would be better to have calls to the DB contained in Banzai rather than directly calling from banzai-web.

cmccully:
We have superseded the need for this in banzai-web which originally drove this PR.

PR #221, opened 2019-07-19
Add extra error handling to astrometry stage

mgdaily:
This added `try...except` clause guards against some of the more exotic HTTP errors coming from the astrometry service, which do not contain any JSON in the response. When attempting to log an error, try to log the JSON, and if an exception is raised, simply log that an error occurred and the image filename.


cmccully:
Can you add a message to the change log.

PR #226, opened 2019-07-29
Bugfix/astrometry requests error handling

mgdaily:
Use a context manager for request to the astrometry service to prevent requests' `ChunkedEncodingError` by forcing the connection to be closed after completion of the stage. 

I also removed some specificity from the exception handling. We check for a `ConnectionError` at first, but any other exception will be handled and appropriately logged. Prior to this, we only handled `HTTPError` exceptions in this block.


mgdaily:
Closing this, as we believe the problem was external to BANZAI. We are not seeing this failure mode occurring anymore.

PR #227, opened 2019-07-30
Feature/enable reduction for standby

mgdaily:
This enables BANZAI to reduce data from cameras in `STANDBY` state. I also added a quick test to check that instruments in STANDBY are marked as SCHEDULABLE in the DB. Updated the ConfigDB test fixture to modify kb98's state from SCHEDULABLE->STANDBY for this purpose.

### DIFF
--- a/banzai/tests/data/configdb_example.json
+++ b/banzai/tests/data/configdb_example.json
@@ -555,7 +555,7 @@
               "instrument_set": [
                 {
                   "id": 66,
-                  "state": "SCHEDULABLE",
+                  "state": "STANDBY",
                   "telescope": "http://configdb.lco.gtn/telescopes/22/?format=json",
                   "science_camera": {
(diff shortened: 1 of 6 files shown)


banzai/tests/data/configdb_example.json
```diff
@@ -555,7 +555,7 @@
               "instrument_set": [
                 {
                   "id": 66,
-                  "state": "SCHEDULABLE",
+                  "state": "STANDBY",
```
dmcollom: This technically does what you want without breaking end to end tests but I'm not a big fan.

mgdaily: For the record, David and I had a conversation about this. He thought I was changing all schedulable instruments to standby. That's not the case - just changing kb98. The goal of this is to replicate a realistic ConfigDB configuration.

PR #228, opened 2019-08-01
Adapt to observation portal

xuiris:
Collab with @Ara-yjx 

Change all structure to match the new observation portal.
Update test data.


Ara-yjx:
Ready for re-review.

The key is to set `--max-tasks-per-child 1` for celery. It has almost no negative effect to running speed.
In addition, in celery task, using `@app.task(autoretry_for=(Exception,))` can restart the task when it gets SIGKILL. But this takes more time because tasks might run multiple times.

PR #230, opened 2019-09-09
Remove schedulability logic

mgdaily:
…lity from context and command line. Remove schedulability checks.


cmccully:
Fixed in #231

PR #231, opened 2019-09-12
Significant updates to the architecture

cmccully:
This is meant to make it easier to memory map the data and make the BANZAI as a library easier to use for projects like NRES.


banzai/images.py
```diff
@@ -1,272 +1,766 @@
...
+        del self.uncertainty
+
+    def __isub__(self, value):
+        if isinstance(value, CCDData):
+            self.data -= value.data
+            self.uncertainty = np.sqrt(value.uncertainty * value.uncertainty + self.uncertainty * self.uncertainty)
+            self.mask |= value.mask
+        else:
+            self.data -= value
+        return self
```
gmbrandt: Do both __isub__ and __sub__ need to be defined for image? If a and b were CCDData In a little example I did, it seems like __sub__ is sufficient for getting both x = a-b  to work and a-=b to work. Or is __isub__ for something other than defining a-=b ?

banzai/images.py
```diff
@@ -1,272 +1,766 @@
...
+    def __imul__(self, value):
+        self.data *= value
+        self.meta['SATURATE'] *= value
+        self.meta['GAIN'] *= value
+        self.meta['MAXLIN'] *= value
+        return self
+
+    def __idiv__(self, other):
+        self.__imul__(1.0 / other)
+        return self
```
gmbrandt: I thought __idiv__ and __div__ are replaced with __itruediv__ and __truediv__ in python 3. (https://docs.python.org/3/library/operator.html)

In a little example I did with a simple class, in python 3 I can define custom division using __truediv__ but __div__ does not work because it is depricated. How did you get around this?

banzai/bias.py
```diff
@@ -35,28 +31,11 @@ def calibration_type(self):
...
-        logging_tags.update(master_logging_tags)
-
-        logger.info('Subtracting bias', image=image,  extra_tags=logging_tags)
-
+        image -= master_calibration_image.bias_level
+        image.meta['BIASLVL'] = master_calibration_image.bias_level, 'Bias level that was removed after overscan'
+        image -= master_calibration_image
+        image.meta['L1IDBIAS'] = master_calibration_image.filename, 'ID of bias frame'
+        image.meta['L1STATBI'] = 1, "Status flag for bias frame correction"
         return image
```
gmbrandt: I really like this.

banzai/images.py
```diff
@@ -1,272 +1,766 @@
...
+        self._validate_mask(mask)
+        self.mask = self._init_array(mask)
 
-    def add_column(self, arr, name, index=None):
-        self._data_table.add_column(Column(arr), name=name, index=index)
+    def __del__(self):
+        for handle in self._file_handles:
+            handle.close()
+        del self.data
+        del self.mask
```
gmbrandt: Do we need to override the standard delete method? When we do del CCDdata_instance, does it not delete every attribute it has from memory?

banzai/images.py
```diff
@@ -1,272 +1,766 @@
...
+    def append(self, hdu):
+        self._hdus.append(hdu)
+
+    def remove(self, hdu):
+        self._hdus.remove(hdu)
+
+    def insert(self, index, hdu):
+        self._hdus.insert(index, hdu)
+
+    def subtract(self, value, kind=None):
```
gmbrandt: will this be replaced by ` __sub__` as well? I find using `__sub__` very clean.

banzai/images.py
```diff
@@ -1,272 +1,766 @@
...
+            x_binning, y_binning = hdu.binning
+            x_binnings.append(x_binning)
+            y_binnings.append(y_binning)
+        return [min(x_binnings), min(y_binnings)]
+
+    def __sub__(self, other):
+        return self.primary_hdu - other.primary_hdu
+
+    def __isub__(self, other):
+        self.primary_hdu.__isub__(other)
```
gmbrandt: was this supposed to be `.__sub__(other)`?

banzai/dbs.py
```diff
@@ -294,99 +213,23 @@ def add_or_update_record(db_session, table_model, equivalence_criteria, record_a
     return record
 
 
-def populate_calibration_table_with_bpms(directory, runtime_context):
```
mgdaily: Do you have another way to add BPMs to the calibration table now? The reason I ask is because when commissioning a new instrument, we use this functionality to add in a BPM that's dropped into `/cache` by someone like Nikolaus.

See: https://sites.google.com/lco.global/docs/data-pipelines/banzai-data-pipeline?authuser=0#h.p_5pyUK_Txlq8V

setup.cfg
```diff
@@ -53,7 +100,7 @@ edit_on_github = True
 github_project = lcogt/banzai
 
 # version should be PEP440 compatible (http://www.python.org/dev/peps/pep-0440)
-version = 0.28.5
+version = 0.29.0
 
 [options]
 setup_requires =
```
mgdaily: Should we specify (and pin) dependencies here, rather than in the dockerfile?

banzai/dbs.py
cmccully: Yes. There is a command line script that does this.

setup.cfg
cmccully: I think we should just specifiy them here an not pin then unless there is specific need.

cmccully: Don't we already specify them here?

PR #233, opened 2019-10-19
Feature/obs portal

mgdaily:
This is a cherry-picked and cleaned up branch containing Ara and Iris' changes to BANZAI for the obs portal.

@Ara-yjx @xuiris thanks for the help on this! It's much appreciated. :)

A note when deploying this: we'll need to switch over the `LAKE_URL` rancher environment variable to `OBSERVATION_PORTAL_URL`


banzai/celery.py
```diff
@@ -86,10 +86,13 @@ def stack_calibrations(self, min_date: str, max_date: str, instrument_id: int, f
...
-    for block in blocks:
-        for molecule in block['molecules']:
-            if frame_type.upper() == molecule['type']:
-                expected_image_count += molecule['exposure_count']
+    for observation in observations:
+        for configuration in observation['request']['configurations']:
+            if frame_type.upper() == configuration['type']:
+                for instrument_config in configuration['instrument_configs']:
+                    expected_image_count += instrument_config['exposure_count']
+    logger.info('expected image count: {0}'.format(str(expected_image_count)))
```
dmcollom: Suuuuuuuper nitpicky but let's combine these into a single logging statement, because perusing logs is terrible when we're effectively using these two statements to just compare two numbers.

mgdaily: Good call.

PR #240, opened 2020-01-16
Feature/s3 migration

mgdaily:
This is the first pass at outfitting BANZAI to use the Archive API and the ingester library to reduce images while listening to the `archived_fits` queue.

Things of note:
`process_image` and the opening of images now operate on queue messages, and have been modified to be able to pull files from the archive API if they are not present locally.

The saving of reduced images is largely the same - they are posted to the ingester, and an frameid is returned. This frameid is then written into the db record for the image, so it may be retrieved from the archive API later if it is no longer present on disk. They are then written to disk as before, as we have decided to cache certain frames to avoid exorbitant data transfer costs.


banzai/calibrations.py
```diff
@@ -108,37 +108,48 @@ def master_selection_criteria(self):
...
+        if master_calibration is None:
             self.on_missing_master_calibration(image)
             return image
 
-        master_calibration_image = FRAME_CLASS(self.runtime_context, filename=master_calibration_filename)
+        file_info = self.construct_file_info_for_master_calibration(master_calibration)
+
+        master_calibration_image = FRAME_CLASS(self.runtime_context, file_info=file_info)
+        # cache the master calibration image for future use
+        if not os.path.isfile(file_info.get('path')):
```
cmccully: I would just put this logic in the constructor rather than here.

banzai/calibrations.py
```diff
@@ -225,8 +236,10 @@ def create_master_calibration_header(old_header, images):
     return header
 
 
-def run_master_maker(image_path_list, runtime_context, frame_type):
-    images = [image_utils.read_image(image_path, runtime_context) for image_path in image_path_list]
+def run_master_maker(calibration_image_records, runtime_context, frame_type):
+    file_info_list= [{'path': os.path.join(record.filepath, record.filename),
+                      'frameid': record.frameid} for record in calibration_image_records]
+    images = [image_utils.read_image(file_info, runtime_context) for file_info in file_info_list]
```
cmccully: The caching logic may be even better to have in here.

banzai/images.py
```diff
@@ -53,12 +52,14 @@ def __init__(self, runtime_context, filename=None, data=None, data_tables=None,
         if extension_headers is None:
             extension_headers = []
 
-        if filename is not None:
-            data, header, bpm, extension_headers = fits_utils.open_image(filename)
+        if file_info is not None:
+            data, header, bpm, extension_headers = fits_utils.open_image(file_info, runtime_context)
```
cmccully: Caching here.

banzai/utils/file_utils.py
```diff
@@ -85,4 +132,31 @@ def get_calibration_filename(image):
...
         return cal_file
+
     return get_calibration_filename
+
+
+def get_basename(path):
+    basename = None
+    if path is not None:
+        filename = os.path.basename(path)
+        if filename.find('.') > 0:
```
cmccully: do a replace('.fits', '').replace('.fz', '')

banzai/utils/file_utils.py
```diff
@@ -85,4 +132,31 @@ def get_calibration_filename(image):
...
+    return basename
+
+
+def download_from_s3(file_info, output_directory, runtime_context):
+    frame_id = file_info.get('frameid')
+    if frame_id is not None:
+        url = f'{runtime_context.ARCHIVE_FRAME_URL}/{frame_id}'
+    else:
+        basename = get_basename(file_info.get('path'))
+        url = f'{runtime_context.ARCHIVE_FRAME_URL}/?basename={basename}'
```
cmccully: I think this returns a list instead of a single element so make sure we cover that case.

banzai/utils/fits_utils.py
```diff
@@ -119,38 +121,40 @@ def open_fits_file(filename):
...
+
     if file_extension == '.fz':
         with tempfile.TemporaryDirectory() as tmpdirname:
-            output_filename = os.path.join(tmpdirname, base_filename)
-            os.system('funpack -O {0} {1}'.format(output_filename, filename))
-            hdulist = fits.open(output_filename, 'readonly')
+            output_filepath = os.path.join(tmpdirname, base_filename)
+            if need_to_download:
+                downloaded_filepath = download_from_s3(file_info, tmpdirname, runtime_context)
+                os.system('funpack -O {0} {1}'.format(output_filepath, downloaded_filepath))
```
cmccully: Shouldn't we just make the downloaded path the same as the usual path?

banzai/utils/fits_utils.py
```diff
@@ -233,3 +237,69 @@ def get_configuration_mode(header):
 
     header['CONFMODE'] = configuration_mode
     return configuration_mode
+
+
+def get_primary_header(file_info, runtime_context):
+    try:
+        hdulist = open_fits_file(file_info, runtime_context)
+        return hdulist[0].header
```
cmccully: Double check that open_fits_file deals with fpacked files correctly and does the rearranging correctly.

banzai/dbs.py
```diff
@@ -403,6 +400,9 @@ def save_calibration_info(output_file, image, db_address=_DEFAULT_DB):
                              'is_master': image.is_master,
                              'is_bad': image.is_bad,
                              'attributes': {}}
+        if image.file_info is not None:
+            # TODO: If frameid is missing, how should we handle that?
+            record_attributes['frameid'] = image.file_info.get('frameid', 0)
```
mgdaily: Fix this TODO.

banzai/utils/file_utils.py
```diff
@@ -16,6 +23,46 @@ def post_to_archive_queue(image_path, broker_url, exchange_name='fits_files'):
...
+                retry = True
+                try_counter += 1
+            except BackoffRetryError as exc:
+                logger.debug('BackoffRetry Exception occured: {0}. Retrying.'.format(exc),
+                             extra_tags={'filename': filepath})
+                if try_counter > 5:
+                    logger.warning('Giving up because we tried too many times.', extra_tags={'filename': filepath})
+                    retry = False
+                else:
+                    sleep(5 ** try_counter)
```
Fingel: Is it possible to use celery or something for this? That way you could schedule the task to execute later instead of blocking. https://github.com/LCOGT/ingester/blob/6a4eb8f4de65566a931eecaff84f43e8fca46346/tasks.py#L61

mgdaily: This could definitely be a celery task. Good call.

banzai/utils/fits_utils.py
mgdaily: This does indeed do what we expect.

banzai/utils/fits_utils.py
mgdaily: `downloaded_filepath` is the `temp_dir/{filename}`, which `funpack` will operate on. This is returned by the `download_from_s3` method as the path to where the file was downloaded to. `output_filepath` is where `funpack` will write the unpacked file to.

In both cases, `funpack` is writing the output file (`.fits` - funpacked) to the `output_filepath`

banzai/calibrations.py
mgdaily: We have decided to keep the caching logic here, as we do not wish to cache every frame that is downloaded.

The reason for placing the caching logic in the `ApplyCalibration` class is that every time an `ApplyCalibration` stage is run (any stage that utilizes a master calibration image to perform reduction), the master cal is re-opened. When working on `/archive` this wasn't a big deal, but now with the likelihood of downloading the file from s3, we need to be smarter about this as we don't wish to re-download a master calibration image 3-4 times or more per reduction.

banzai/calibrations.py
mgdaily: See comment above.

banzai/images.py
mgdaily: See comment above.

banzai/settings.py
```diff
@@ -78,5 +78,10 @@
 
 TELESCOPE_FILENAME_FUNCTION = 'banzai.utils.file_utils.telescope_to_filename'
 
+ARCHIVE_API_ROOT = os.getenv('API_ROOT', 'https://archive-api.lco.global/frames')
```
eheinrich: Are these environment variables not going to be set? It might be safer to default to non-production values.

banzai/utils/file_utils.py
```diff
@@ -85,4 +131,18 @@ def get_calibration_filename(image):
...
     return get_calibration_filename
+
+
+def download_from_s3(file_info, output_directory, runtime_context):
+    frame_id = file_info.get('frameid')
+    logger.info(f"Downloading file {file_info.get('filename')} from archive. ID: {frame_id}.",
+                extra_tags={'filename': file_info.get('filename')})
+    url = f'{runtime_context.ARCHIVE_FRAME_URL}/{frame_id}'
+
+    response = requests.get(url, headers=runtime_context.ARCHIVE_AUTH_TOKEN, stream=True).json()
```
eheinrich: What is the stream=True here actually getting you? Parameter description is "if False, the response content will be immediately downloaded." If it's True then the download is delayed until the content is accessed, but in this case the content is being accessed right away https://requests.readthedocs.io/en/master/api/#main-interface

mgdaily: You're absolutely right. Good catch.

banzai/settings.py
mgdaily: Removed the default. We are always going to set these.

PR #243, opened 2020-02-10
Add retry+exponential backoff to s3 download.

mgdaily:
We've been seeing some ChunkedEncodingErrors when pulling frames directly from s3, so I've added some basic retry logic with exponential backoff to the s3 downloading function. I found this cool library Tenacity (https://github.com/jd/tenacity) over the weekend that provides a nice, simple interface for setting up retry logic.

Also added the `stream=True` flag to the `requests.get()` call to s3.

For now, I have the parameters set in the decorator; there probably is a better place to put them?

### DIFF
--- a/banzai/utils/fits_utils.py
+++ b/banzai/utils/fits_utils.py
@@ -10,6 +10,7 @@ import numpy as np
 from astropy.io import fits
 from astropy.coordinates import SkyCoord
 from astropy import units
+from tenacity import retry, wait_exponential, stop_after_attempt
 
 logger = logging.getLogger('banzai')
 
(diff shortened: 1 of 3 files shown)


banzai/utils/fits_utils.py
```diff
@@ -213,26 +214,28 @@ def get_basename(path):
     return basename
 
 
+@retry(wait=wait_exponential(multiplier=2, min=4, max=10), stop=stop_after_attempt(4), reraise=True)
```
mgdaily: I realized these parameters aren't obvious, so I'll throw a comment in.

Stop after 4 attempts, and back off exponentially with a minimum wait time of 4 seconds, and a maximum of 10. If it fails, "reraise" the original exception back up to the caller.

mgdaily: https://tenacity.readthedocs.io/en/latest/index.html#waiting-before-retrying

PR #247, opened 2020-02-27
proper uncertainty propagation on divide.

gmbrandt:
This PR fixes error propagation between two instances of `CCDData` when they are divided, using this eq 13 [here](http://ipl.physics.harvard.edu/wp-uploads/2013/03/PS3_Error_Propagation_sp13.pdf).

### DIFF
--- a/banzai/data.py
+++ b/banzai/data.py
@@ -123,12 +123,14 @@ class CCDData(Data):
         return self
 
     def __itruediv__(self, value):
-        if isinstance(value, CCDData):
-            self.data /= value.data
-            self.uncertainty /= value.uncertainty
+        if isinstance(value, type(self)):
+            self.uncertainty = np.abs(self.data) * np.sqrt(
+                (self.uncertainty / self.data) ** 2 + (value.uncertainty / value.data) ** 2)
+            self.data = np.divide(self.data, value.data)
             self.meta['SATURATE'] /= value.meta['SATURATE']
             self.meta['GAIN'] /= value.meta['GAIN']
             self.meta['MAXLIN'] /= value.meta['MAXLIN']
+            self.mask |= value.mask
         else:
             self.__imul__(1.0 / value)
         return self
--- a/banzai/data.py
+++ b/banzai/data.py
@@ -123,9 +123,9 @@ class CCDData(Data):
         return self
 
     def __itruediv__(self, value):
-        if isinstance(value, type(self)):
-            self.uncertainty = np.abs(self.data / value.data) * np.sqrt(
-                (self.uncertainty / self.data) ** 2 + (value.uncertainty / value.data) ** 2)
+        if isinstance(value, CCDData):
+            self.uncertainty = np.abs(self.data / value.data) * \
+                               np.sqrt((self.uncertainty / self.data) ** 2 + (value.uncertainty / value.data) ** 2)
             self.data = np.divide(self.data, value.data)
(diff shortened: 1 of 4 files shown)


banzai/data.py
```diff
@@ -123,12 +123,14 @@ def __imul__(self, value):
         return self
 
     def __itruediv__(self, value):
-        if isinstance(value, CCDData):
```
gmbrandt: I am not sure whether this should be CCDData or type(self), it seems to make sense to be type(self) instead of being restricted to the parent class.

cmccully: I disagree actually. Basically if the data object has an attribute uncertatinty we want to propagate it even if one is an image ccd data class  (e.g. from a bias frame) and the other is a spectrum ccddata object.

gmbrandt: Great, thanks for clarifying. I agree now.

banzai/data.py
```diff
@@ -123,12 +123,14 @@ def __imul__(self, value):
...
-            self.data /= value.data
-            self.uncertainty /= value.uncertainty
+        if isinstance(value, type(self)):
+            self.uncertainty = np.abs(self.data / value.data) * np.sqrt(
+                (self.uncertainty / self.data) ** 2 + (value.uncertainty / value.data) ** 2)
```
cmccully: The spacing on this is weird. maybe do 
self.uncertainty = np.abs(self.data / value.data) * \
    np.sqrt((self.uncertainty / self.data) ** 2 + (value.uncertainty / value.data) ** 2)

banzai/tests/utils.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, image_multiplier=1.0, nx=101, ny=103, name='test_image', read
         if meta is not None:
             self.meta = meta
         else:
-            self.meta = Header()
+            self.meta = Header({'SATURATE': 1, 'GAIN': 1, 'MAXLIN': 1})
```
mgdaily: Minor nitpick. If we're not using these header keywords in many other tests, we may just want to specify them when instantiating the `FakeCCDData` object in the test itself, rather than having them automatically hardcoded/populated in the constructor.

banzai/tests/test_frames.py
```diff
@@ -27,6 +27,23 @@ def test_subtract():
...
     assert (test_data.data == 3 * np.ones(test_data.data.shape)).all()
+    assert np.allclose(test_data.uncertainty, 3)
+
+
+def test_uncertainty_propagation_on_divide():
+    data1 = FakeCCDData(data=np.array([2, 2]), uncertainty=np.array([2, 2]))
+    data2 = FakeCCDData(data=np.array([1, 1]), uncertainty=np.array([2, 2]))
+    data1 /= data2
+    assert np.allclose(data1.data, 2)
+    assert np.allclose(data1.uncertainty, 2*np.sqrt(5))
```
mgdaily: Maybe assert that the `SATURATE`, `GAIN` and `MAXLIN` keywords were properly calculated too?

banzai/tests/utils.py
cmccully: Also a saturate keyword of 1 and maxlin of 1 has the potential to break things.

banzai/tests/test_frames.py
gmbrandt: I agree, added now.

PR #250, opened 2020-03-06
Update lco-ingester dependency version

irasnyd:
The lco-ingester library has been updated to version 2.1.14, which
itself contains an update to one of its dependencies,
opentsdb-python-metrics, which publishes metrics data on the standard
HTTP port 80 by default, rather than using a custom port number.

This behavior is also now configurable, rather than being hard-coded
into the library.

With this change, the LCO IT staff will be able to migrate the OpenTSDB
server to our newer hardware platform.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+0.28.8 (2020-03-09)
+-------------------
+- Upgrade ingester lib version to address OpenTSDB default HTTP port issue.
+- Fix EXPTIME to be a float, rounded to 6 digits.
+
 0.28.7 (2020-03-05)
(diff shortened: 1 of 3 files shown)


irasnyd:
I'll let you guys actually merge and deploy this. There is no urgent rush, however this will really help the IT Team significantly in the very near term. We would appreciate if this can get merged and deployed within the next week or so.

mgdaily:
Thank you Ira!

For the purposes of documentation, I have also rolled in a fix to the EXPTIME into this PR, forcing it to be a float rounded to 6 digits.

PR #256, opened 2020-04-01
Plumbed frame id through if we get download a file from s3.

cmccully:


banzai/utils/fits_utils.py
```diff
@@ -122,8 +122,9 @@ def open_fits_file(file_info, context, is_raw_frame=False):
     if file_info.get('path') is not None and os.path.exists(file_info.get('path')):
         buffer = open(file_info.get('path'), 'rb')
         filename = os.path.basename(file_info.get('path'))
+        frame_id = None
     elif file_info.get('frameid') is not None:
-        buffer = download_from_s3(file_info, context, is_raw_frame=is_raw_frame)
+        buffer, frame_id = download_from_s3(file_info, context, is_raw_frame=is_raw_frame)
```
mgdaily, line 127: Could you just directly set `frame_id = file_info.get('frameid')` in this block, rather than returning it from `download_from_s3()`?

cmccully:
Merged into astrometry photometry branch.

PR #258, opened 2020-04-24
Fixes for logging in celery

cmccully:


banzai/celery.py
```diff
@@ -36,6 +38,14 @@ def configure_workers(**kwargs):
     reload(metric_wrappers)
 
 
+app = Celery('tasks')
```
mgdaily, line 41: I don't think you want to initialize a new celery app, here, but just use the existing one that's defined on line 15.

PR #265, opened 2020-05-29
Remove configdb test data that will be removed from configdb

eheinrich:
Data that is being removed from the configdb because it is now represented in another way is removed from the test dump of configdb data.


mgdaily:
@eheinrich Thanks for taking care of this!

PR #266, opened 2020-06-02
Check that extension is in HDUList before trying to re-order. Add test.

mgdaily:


banzai/utils/fits_utils.py
```diff
@@ -221,11 +221,18 @@ def reorder_hdus(hdu_list: fits.HDUList, extensions: list):
...
     for idx, extension_name in enumerate(extensions):
         if hdu_list[idx].name != extension_name:
-            hdu = hdu_list[extension_name]
+            try:
+                hdu = hdu_list[extension_name]
+            except KeyError:
+                # If the extension is not present in the HDUList, do not try and re-order it.
+                logger.warning(f"Extension {extension_name} not present in HDUList.")
+                extensions.remove(extension_name)
+                continue
```
cmccully, line 235: Removing from a list in a for loop is messy. I think I would recommend the following:
def reorder_hdus(hdu_list, extensions):
    if extensions is None:
        extensions = []
    extensions += [hdu.name for hdu in hdu_list]
    # Use an ordered dict to get unique elements
    extensions = list(OrderedDict.fromkeys(extensions))
    hdu_list.sort(key=lambda x: extensions.index(x.name))

PR #269, opened 2020-06-16
Add logic to munge data and trim sections for fs01 into correct form.

mgdaily:
This is to deal with the fact that fs01 has pre-scan turned on, and its detector and data sections are incorrect. Turning off pre-scan and appropriately adjusting the header keywords requires a great deal of data validation and effort in low-level software.

This logic should be removed when the fs01 spectral camera is retired.

### DIFF
--- a/banzai/lco.py
+++ b/banzai/lco.py
@@ -400,7 +400,7 @@ class LCOFrameFactory(FrameFactory):
                         hdu_list += self._munge_data_cube(hdu)
                     # update datasec/trimsec for fs01
                     if hdu.header.get('INSTRUME') == 'fs01':
-                        self._update_spectral_sections(hdu)
+                        self._update_fs01_sections(hdu)
                     if hdu.data.dtype == np.uint16:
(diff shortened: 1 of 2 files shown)


banzai/lco.py
```diff
@@ -485,6 +488,18 @@ def _init_saturate(image):
                          image=image)
             raise MissingSaturate
 
+    @staticmethod
+    def _update_spectral_sections(hdu):
```
cmccully, line 492: I would mark this fs01 in the function name.

banzai/lco.py
```diff
@@ -485,6 +488,18 @@ def _init_saturate(image):
...
+        """
+        Manually update data and trim sections for fs01 spectral camera
+        :param hdu: Astropy ImageHDU
+        """
+        section_keywords = {'TRIMSEC': '[2:2046,3:2015]',
```
cmccully, line 497: You might add a case to only set these values when the trimsec and datasec are their current values.

PR #276, opened 2020-07-09
Feature/multiple science cameras

jnation3406:
The unit tests are passing, but I am unsure how to run the integration test suite for banzai. I had to add a FakeImage class to the test/utils.py to get a few of the unit tests to pass... not sure how that works in the master branch right now \_O_/.


jnation3406:
Configdb-dev currently has the extra structures and multiple science_cameras, but prod does not yet, so don't merge this change in to prod until then.

jnation3406:
Can I copy the configdb_example.json file from this branch into a banzai_nres branch? I think that is the only change needed for banzai nres because it looks like it uses banzai's configdb access code as far as I can tell. It looks like it relies on an older version of banzai (0.28.0) right now, but I assume that can be upgraded to the latest once this change is in banzai?

banzai/lco.py
```diff
@@ -22,7 +22,7 @@ class LCOObservationFrame(ObservationFrame):
     def get_output_filename(self, runtime_context):
         # TODO add a mode for AWS filenames
         output_directory = os.path.join(runtime_context.processed_path, self.instrument.site,
-                                        self.instrument.name, self.epoch, 'processed')
+                                        self.instrument.camera, self.epoch, 'processed')
```
cmccully, line 25: This will break NRES. It may be the right thing to do, but we need to make NRES consistent.

banzai/dbs.py
```diff
@@ -147,18 +147,18 @@ def parse_configdb(configdb_address):
...
-                                      'name': ins.get('code'),
-                                      'type': sci_cam['camera_type']['code']}
-                        # hotfix for configdb
-                        if not instrument['name']:
-                            instrument['name'] = instrument['camera']
-                        if instrument['name'] in CAMERAS_FOR_INSTRUMENTS:
-                            instrument['camera'] = CAMERAS_FOR_INSTRUMENTS[instrument['name']]
-                        instruments.append(instrument)
+                    for sci_cam in ins['science_cameras']:
+                        if sci_cam is not None:
```
cmccully, line 151: Do we still need to check for None here?

banzai/tests/utils.py
```diff
@@ -164,3 +164,7 @@ def __init__(self):
         self.frameid = 1234
         self.filepath = '/tmp/'
         self.filename = 'test.fits'
+
+
+class FakeImage(FakeCalImage):
```
cmccully, line 169: I'm not sure why this is necessary. Hopefully we can remove it?

jnation3406: I added this FakeImage class because one of the test classes (qc_stage_integration_tests.py) imports it, and it did not exist. So I don't know how the current set of unit tests were passing, but this is what I had to do to get them to pass running locally using the command specified in the readme.

banzai/dbs.py
jnation3406: I don't think there is a case where the instrument has science_cameras, but they are each none, so I would think this is unnecessary.

banzai/lco.py
jnation3406: @mgdaily Can you comment on this? If this method is to build the AWS daydir, then I think we store the data by instrument code (instrument.name) rather than camera code (instrument.camera), but I am not sure how this is used exactly?

mgdaily: Yes, that's correct. This builds the daydir structure for when we write to disk. Name is what we want in this case.

banzai/tests/utils.py
mgdaily: I'll take a look at this and update it.

PR #280, opened 2020-07-13
Revert engine creation to known good formula, as DB connections were …

mgdaily:
…not getting cleaned up.

### DIFF
--- a/banzai/dbs.py
+++ b/banzai/dbs.py
@@ -36,10 +36,8 @@ def get_session(db_address):
     session: SQLAlchemy Database Session
     """
     # Build a new engine for each session. This makes things thread safe.
-    engine = create_engine(db_address)
-    # Copy the metadata object to stop engine collisions. This may be not how sqlalchemy is intended to be used.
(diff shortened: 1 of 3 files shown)


banzai/dbs.py
```diff
@@ -36,10 +36,8 @@ def get_session(db_address):
...
     # Build a new engine for each session. This makes things thread safe.
-    engine = create_engine(db_address)
-    # Copy the metadata object to stop engine collisions. This may be not how sqlalchemy is intended to be used.
-    metadata = copy.deepcopy(Base.metadata)
-    metadata.bind = engine
+    engine = create_engine(db_address, poolclass=pool.NullPool)
+    Base.metadata.bind = engine
 
     # We don't use autoflush typically. I have run into issues where SQLAlchemy would try to flush
     # incomplete records causing a crash. None of the queries here are large, so it should be ok.
```
cmccully, line 43: should we add and engine.dispose() to the end of the finally in the context mananger?

mgdaily: let's try that.

PR #287, opened 2020-08-12
Feature/helm deployment

mgdaily:
This adds a dev helm deployment for the BANZAI imaging pipeline. 

The dev stack will be deployed to the LCO k8s dev namespace upon pushes to dev. This application listens to the `banzai_dev_pipeline` queue, which is fed by the standard FITS exchange. Reduced images are pushed to/pulled from the dev archive. Raw images are pulled from the production archive.

The helm chart is very similar to banzai-nres, with the exception that the `initContainers` which run when `useDockerizedDatabase=True` are now configured to run when the listener pod is spun up, rather than the worker pod. This means that when a new worker pod is spun up using the autoscaling feature, we won't repeat the DB init process, since there isn't a need to. This makes for faster scaling performance, which is nice.

Beyond the code changes, a great deal of effort was made to get ALL of the BPMs into the production science archive. This is a prerequisite for getting production BANZAI onto k8s.


Jenkinsfile
```diff
@@ -43,6 +43,23 @@ pipeline {
 				}
 			}
 		}
+		stage('DeployDevStack') {
+			when {
+				anyOf {
+				branch 'feature/helm-deployment'
```
mgdaily, line 49: TODO: Make this happen on `dev` before we merge

PR #290, opened 2020-08-19
AWS Deployment

mgdaily:
This PR adds a CronJob to the BANZAI deployment which updates the instruments table every 5 minutes, just like we have currently in Rancher. I also changed some of the names of the values to agree with the camelCase standard, defined by Helm. https://helm.sh/docs/chart_best_practices/values/#naming-conventions

* Adds readiness probes to e2e test pod to address intermittent failures
* Removes automatic dev deployment 
* Configures working directory for celerybeat to be `/tmp`, as we have configured our production deployment with a large emptyDir volume at `/tmp`


Jenkinsfile
```diff
@@ -77,6 +60,8 @@ pipeline {
 					sh('kubectl --kubeconfig=${KUBERNETES_CREDS} -n dev apply -f banzai/tests/e2e-k8s.yaml')
 					// wait for the test pod to be running
 					sh('kubectl --kubeconfig=${KUBERNETES_CREDS} -n dev wait --for=condition=Ready --timeout=60m pod/banzai-e2e-test')
+					// sleep for an additional 30 and allow all containers to get into the proper state
+					sh('sleep 30s')
```
cmccully, line 64: Is this necessary? This should be taken care of in the wait command above. If not, we should make sense the containers are not set to ready until they can be used.

helm-chart/banzai/requirements.lock
```diff
@@ -0,0 +1,9 @@
+dependencies:
+- name: postgresql
+  repository: https://kubernetes-charts.storage.googleapis.com
```
cmccully, line 3: Does this address Ira's recent comments that the locations of charts has moved?

helm-chart/banzai/templates/_helpers.tpl
```diff
@@ -87,62 +90,63 @@ Define shared environment variables
...
-    secretKeyRef:
-      name: banzai-secrets
-      key: AWS_ACCESS_KEY_ID
+  value: {{ .Values.awsAccessKeyId | quote }}
 - name: AWS_SECRET_ACCESS_KEY
-  valueFrom:
-    secretKeyRef:
-      name: banzai-secrets
-      key: AWS_SECRET_ACCESS_KEY
+  value: {{ .Values.awsSecretAccessKey | quote }}
```
cmccully, line 132: Why did we move away from using keys?

helm-chart/banzai/values-dev.yaml
```diff
@@ -36,34 +36,38 @@ tolerations: []
 
 affinity: {}
 
-CALIBRATE_PROPOSAL_ID: calibrate
+calibrateProposalId: calibrate
```
cmccully, line 39: Why did we switch to camel case instead of env variables?

mgdaily: See the description. These are values files, and camelcase is the standard.

helm-chart/banzai/requirements.lock
mgdaily: Fixed this.

Jenkinsfile
mgdaily: Added readiness probes to the rabbitmq, redis, and celery containers to make sure we are ready for test invocation to avoid intermittent failures.

helm-chart/banzai/templates/_helpers.tpl
mgdaily: Discussed over Zoom. It may make sense to re-visit this when and if we decide if we want to provide a helm chart to the public.

cmccully (approved):
So after a lot of discussion, we are going to merge this and we will address some of these comments in future work.

PR #293, opened 2020-09-10
Make ra/dec parsing an overrideable abstract method for ObservationFr…

mgdaily:
…ame class.


mgdaily:
Closed since Curtis fixed this in downstream.

PR #296, opened 2020-10-26
Fix/remove astropy helpers

mgdaily:
This PR removes BANZAI's dependence on astropy-helpers for building and packaging.

* Updates testing on Jenkins and Travis
* Adds a full `tox.ini` for ease of testing using multiple python, astropy, numpy versions. tox is also used to build docs and perform pycodestyle checks (see travis pipeline for examples)
* Updated the README with updated installation and testing instructions. 

All of these changes are pretty transparent. I don't believe BANZAI-NRES will have any issues when installing upstream BANZAI via pip. pip installation of packages, even local ones, is now the preferred method over `python setup.py install`.

I also ran into an issue with Travis, since BANZAI was on the old travis-ci.org. This was causing builds to take over an hour, since they are migrating resources. I've migrated BANZAI to use travis-ci.com. Again, this is pretty transparent. We should probably switch BANZAI-NRES over to travis-ci.com as well.


mgdaily:
Closing to force Travis into building on branch state.

setup.cfg
```diff
@@ -103,37 +100,45 @@ github_project = lcogt/banzai
...
     logutils
-    numpy==1.17.4
+    numpy
     cython
     mysql-connector-python
     lcogt_logging==0.3.2
     sep
     kombu==4.4.0
-    amqp==2.4.2
+    amqp==2.6.0
```
mgdaily, line 116: Upgraded due to a transitive dependency issue with celery/kombu https://github.com/celery/py-amqp/issues/340

README.rst
```diff
@@ -27,11 +27,11 @@ Please cite the following DOI if you are using processed LCOGT data.
 
 Installation
 ------------
-BANZAI can be installed in the usual way, by running
+BANZAI can be installed using pip, by running from the top-level directory containing `setup.py`.
 
 .. code-block:: bash
-
-    python setup.py install
+    pip install --upgrade pip
```
cmccully, line 33: This feels like bad form to have this in the docs.

mgdaily: Add to setup_requires in setup.cfg

cmccully (approved):
This is approved assuming we can do something sensible with the pip version dependency.

PR #298, opened 2020-10-30
Fixed handling of the saturate keyword when we multiply be an array.

cmccully:


captain-exoplanet (approved):
Looks good, my one comment need not prevent merging. Approved.

mgdaily (commented):
Looks reasonable. I think it would be a good idea to make sure we have good unit test coverage for these new cases, just to be sure that they do what we expect.

PR #300, opened 2020-12-14
Improve and Document Public Helm Chart

mgdaily:
This PR updates and documents BANZAI-imaging's helm chart.

* Pull sensitive values (API keys, passwords, etc...) from kubernetes secrets
* Provide examples of development and production values files.
* Provide deployment instructions, secret creation instructions, and documentation on every configurable value.
* Fixes a bug where the `RAW_DATA_AUTH_TOKEN` would not initialize properly if not explicitly set. Rename some settings variables for clarity.

### DIFF
--- a/banzai/settings.py
+++ b/banzai/settings.py
@@ -90,15 +90,19 @@ TELESCOPE_FILENAME_FUNCTION = 'banzai.utils.file_utils.telescope_to_filename'
 OBSERVATION_PORTAL_URL = os.getenv('OBSERVATION_PORTAL_URL',
                                    'http://internal-observation-portal.lco.gtn/api/observations/')
 
+# Specify different sources of raw and processed data, if needed.
 ARCHIVE_API_ROOT = os.getenv('API_ROOT')
+ARCHIVE_AUTH_TOKEN = os.getenv('AUTH_TOKEN')
 ARCHIVE_FRAME_URL = f'{ARCHIVE_API_ROOT}frames'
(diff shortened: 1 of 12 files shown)


banzai/settings.py
```diff
@@ -90,15 +90,19 @@
...
+# Specify different sources of raw and processed data, if needed.
 ARCHIVE_API_ROOT = os.getenv('API_ROOT')
+ARCHIVE_AUTH_TOKEN = os.getenv('AUTH_TOKEN')
 ARCHIVE_FRAME_URL = f'{ARCHIVE_API_ROOT}frames'
-ARCHIVE_AUTH_TOKEN = {'Authorization': f'Token {os.getenv("AUTH_TOKEN")}'}
-FITS_EXCHANGE = os.getenv('FITS_EXCHANGE', 'archived_fits')
+ARCHIVE_AUTH_HEADER = {'Authorization': f'Token {ARCHIVE_AUTH_TOKEN}'}
 
-RAW_DATA_FRAME_URL = os.getenv('RAW_DATA_FRAME_URL', ARCHIVE_FRAME_URL)
-RAW_DATA_AUTH_TOKEN = {'Authorization' : f'Token {os.getenv("RAW_DATA_AUTH_TOKEN", ARCHIVE_AUTH_TOKEN)}'}
```
mgdaily, line 99: For reference, this is where the auth header bug was. If `RAW_DATA_AUTH_TOKEN` was not set, it would default to `ARCHIVE_AUTH_TOKEN`, which is actually in the format of a JSON auth header. The naming was confusing, so I attempted to clear some things up.

cmccully (approved):

PR #301, opened 2021-01-14
Update to use OCS Ingester 2.2.5

mgdaily:
This ingester version includes updates related to the deletion of the "old" LCO s3 bucket. No functional changes.

### DIFF
--- a/helm-chart/banzai/Chart.yaml
+++ b/helm-chart/banzai/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "1.1.1"
+appVersion: "1.1.2"
 description: A Helm chart to deploy the BANZAI pipeline
 name: banzai
-version: 1.1.0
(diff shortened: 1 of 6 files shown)


helm-chart/banzai/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "1.1.1"
+appVersion: "1.1.2"
 description: A Helm chart to deploy the BANZAI pipeline
 name: banzai
-version: 1.1.0
+version: 1.1.1
```
cmccully, line 5: I dislike that these versions are not in lock step.

mgdaily: Made the helm chart version 1.1.2 to match the application version. Will keep them in lockstep from now on.

cmccully (approved):

PR #304, opened 2021-03-09
Various fixes

mgdaily:
* Fixes a helm chart doc issue
* Removes hardcoded version from setup.cfg
* Places 'not e2e' marker in tox config rather than in GH workflow, a la banzai-nres


.github/workflows/build.yml
```diff
@@ -61,4 +61,4 @@ jobs:
       if: "endsWith(matrix.toxenv, 'build_docs')"
       run: sudo apt-get -y install graphviz
     - name: Run tests
-      run: tox $TOXARGS -- $TOXPOSARGS -m 'not e2e'
```
cmccully, line 64: Is this fixed by #306?

mgdaily: Yes. You and I had the same idea ;)

cmccully (approved):

PR #309, opened 2021-04-19
Added infrastructure to split data products downstream in banzai-nres

cmccully:


banzai/frames.py
```diff
@@ -218,12 +197,21 @@ def __itruediv__(self, other):
             self.primary_hdu.__itruediv__(other)
         return self
 
+    def __getitem__(self, item):
+        return self._hdus[self._hdu_keys[item]]
+
+    def __setitem__(self, key, value):
```
mgdaily, line 203: Should we enforce that whatever the HDU key is matches the Data object's `name` attribute? Or set it here if not specified?

e.g. if a user of this class did something like:

```
image['SCI'] = CCDData(data=data, meta=meta, name='BPM') 
```

we could get into a weird state where the name of the HDU doesn't match the key created

banzai/frames.py
```diff
@@ -218,12 +197,21 @@ def __itruediv__(self, other):
...
+        if not key in self._hdu_keys:
+            self._hdus.append(value)
+            i = len(self._hdus) - 1
+            self._hdu_keys[key] = i
+            self._hdu_keys[self._hdus[-1].name] = i
+        else:
+            self._hdus.remove(self._hdu_keys[key])
```
mgdaily, line 210: I think you want a `del self._hdus[self._hdu_keys[key]]` since you are removing by index. [remove() removes by value](https://www.programiz.com/python-programming/methods/list/remove)

banzai/frames.py
```diff
@@ -21,6 +19,7 @@ def __init__(self, hdu_list: list, file_path: str, frame_id: int = None, hdu_ord
         self.instrument = None
         self.frame_id = frame_id
         self.hdu_order = hdu_order
+        self._hdu_keys = {hdu.name: i for i, hdu in enumerate(hdu_list)}
```
mgdaily, line 22: The name `_hdu_keys` threw me for a bit of a loop. `_extension_indices`, `_hdu_indices` maybe? Something to reflect that we're storing a mapping of HDU name to index.

mgdaily (approved):
Looks good!

PR #312, opened 2021-05-12
Feature/specify celery queue

mgdaily:
Allow for a configurable task queue name on a single RabbitMQ broker, so that we may address every pipeline separately for tools like BANZAI-Web.


banzai/celery.py
```diff
@@ -36,6 +37,14 @@ def configure_workers(**kwargs):
 app = Celery('banzai')
 app.config_from_object('banzai.celeryconfig')
 app.conf.update(broker_url=os.getenv('TASK_HOST', 'redis://localhost:6379/0'))
+celery_task_queue_name = os.getenv('CELERY_TASK_QUEUE_NAME', 'celery')
```
mgdaily, line 40: TODO: Add link to celery docs to explain this config

banzai/tests/test_end_to_end.py
```diff
@@ -45,6 +45,7 @@ def celery_join():
     log_counter = 0
     while True:
         queues = [celery_inspector.active(), celery_inspector.scheduled(), celery_inspector.reserved()]
+        logger.error(f"Queues: {queues}")
```
mgdaily, line 48: TODO: Remove log message

cmccully (approved):

PR #314, opened 2021-05-26
Fix/reduce experimental

mgdaily:
Reduce experimental frames


helm-chart/banzai/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "1.3.1"
+appVersion: "1.3.3"
```
mgdaily, line 2: Forgot to update this for 1.3.2, so that's why there's 2 patch version jumps

cmccully (approved):

PR #316, opened 2021-06-15
Added a tolerance to run on a dedicated node for stacking.

cmccully:


banzai/celery.py
```diff
@@ -116,7 +116,8 @@ def schedule_calibration_stacking(site: str, runtime_context: dict, min_date=Non
                                             'instrument': instrument.camera, 'frame_type': frame_type})
                     stack_calibrations.apply_async(args=(stacking_min_date, stacking_max_date, instrument.id, frame_type,
                                                          vars(runtime_context), blocks_for_calibration),
-                                                   countdown=message_delay_in_seconds)
+                                                   countdown=message_delay_in_seconds,
+                                                   queue=runtime_context.CELERY_STACK_QUEUE_NAME)
```
mgdaily, line 120: This queue also needs to be added to the app.conf

https://docs.celeryproject.org/en/stable/userguide/routing.html#manual-routing

PR #317, opened 2021-08-16
Added zeropoint determination for griz filters.

cmccully:


banzai/tests/test_photometry.py
```diff
@@ -0,0 +1,23 @@
...
+    startdir = os.path.dirname(os.path.abspath(__file__))
+
+    do_photometrytestonFile(f"{startdir}/data/cpt1m012-fa06-20200113-0102-e91.fits.fz", 23.1,-0.008, tmpdir=tmpdir)
+    do_photometrytestonFile(f"{startdir}/data/ogg2m001-ep04-20201006-0097-e91.fits.fz", 25.15, 0.063, tmpdir=tmpdir)
+
+def test_match_catalog():
+    pass
+
+def test_zeropoint():
+    pass
```
mgdaily, line 23: Missing newline

banzai/photometry.py
```diff
@@ -241,3 +243,39 @@ def do_stage(self, image):
...
+    def do_stage(self, image):
+        if image.filter not in ['gp', 'rp', 'ip', 'zs']:
+            return image
+
+        try:
+            # Get the sources in the frame
+            reference_catalog = get_reference_sources(image.meta, self.runtime_context.REFERENCE_CATALOG_URL)
+        except HTTPError as e:
+            logger.error(f'Error retrieving photometric reference catalog: {e}', image=image)
+            return image
```
mgdaily, line 263: Do we want to populate a header value to indicate success/failure? Kind of like WCSERR?

banzai/utils/photometry_utils.py
```diff
@@ -0,0 +1,130 @@
...
+def log_zeropoint_likelihood(theta, mags, mag_errors, catalog_mags, catalog_errors, colors, color_errors):
+    # Our model assumes a zeropoint, color term, and a scatter term
+    # catalog = -2.5 * log10(counts / exptime) + zeropoint + color_term * colors
+    zeropoint, color_term, scatter = theta
+    model = mags + zeropoint + color_term * colors
+    sigma_squared = scatter ** 2.0 + mag_errors ** 2.0 + catalog_errors ** 2.0 + color_errors ** 2.0
+    return -0.5 * np.sum((catalog_mags - model) ** 2.0 / sigma_squared + np.log(2.0 * np.pi * sigma_squared))
+
+
+def fit_photometry(matched_catalog, image_filter, color_to_fit, exptime):
```
mgdaily, line 66: This may warrant a review from Daniel - I don't feel like I have the domain knowledge to say if this is correct or not. But I am trusting that it is.

banzai/photometry.py
```diff
@@ -241,3 +243,39 @@ def do_stage(self, image):
...
+
+
+class PhotometricCalibrator(Stage):
+    color_to_fit = {'gp': 'g-r', 'rp': 'r-i', 'ip': 'r-i', 'zs': 'i-z'}
+
+    def __init__(self, runtime_context):
+        super(PhotometricCalibrator, self).__init__(runtime_context)
+
+    def do_stage(self, image):
+        if image.filter not in ['gp', 'rp', 'ip', 'zs']:
```
mgdaily, line 255: May help to add a log here: "Image taken in a filter that will not be photometrically calibrated" or something?

banzai/settings.py
```diff
@@ -143,3 +144,5 @@
 LOSSLESS_EXTENSIONS = []
 
 CELERY_TASK_QUEUE_NAME = os.getenv('CELERY_TASK_QUEUE_NAME', 'celery')
+
+REFERENCE_CATALOG_URL = os.getenv('REFERENCE_CATALOG_URL', 'http://phot-catalog.lco.gtn/')
```
mgdaily, line 148: Will need to add this to the helm chart

banzai/tests/test_photometry.py
```diff
@@ -0,0 +1,23 @@
...
+
+def test_photcalibration(tmpdir):
+
+    print ("Data are stored in temp dir: ", tmpdir )
+    startdir = os.path.dirname(os.path.abspath(__file__))
+
+    do_photometrytestonFile(f"{startdir}/data/cpt1m012-fa06-20200113-0102-e91.fits.fz", 23.1,-0.008, tmpdir=tmpdir)
+    do_photometrytestonFile(f"{startdir}/data/ogg2m001-ep04-20201006-0097-e91.fits.fz", 25.15, 0.063, tmpdir=tmpdir)
+
+def test_match_catalog():
```
mgdaily, line 19: May want to remove these tests if we're not intending to fill them out.

mgdaily (commented):
Mostly looks good. Had some comments/questions.

banzai/photometry.py
drhaz: There should be some documentation on why the color to be fitted is dependent on the filter.  it is a design choice, but since we are fitting against a catalog color, not instrumental colors were you are limited to what you observed, there is a real choice to me made.  The PanSTARRS to SDSS conversion is entirely based on g-i I think, and they did that fro a good reason.

banzai/tests/test_photometry.py
```diff
@@ -0,0 +1,60 @@
...
+        position_angle = np.random.uniform(0, 360) * units.deg
+        separation = np.random.uniform(0.0, 0.3) * units.arcsec
+        shifted_coordinate = coordinate.directional_offset_by(position_angle, separation)
+        row['ra'], row['dec'] = shifted_coordinate.ra.value, shifted_coordinate.dec.value
+
+    # make sure the overlapping sources are the same as we would expect
+    matched_catalog = photometry_utils.match_catalogs(input_catalog, reference_catalog)
+    for i in range(100):
+        assert i in matched_catalog['source_id']
+    assert len(matched_catalog) == 100
```
drhaz, line 47: I think this test can fail, but with a low probability, if the inserted spurious detection is closer to the refernce target than the randomly spereated original source. Low probability, but this is not iron-clad.

banzai/utils/photometry_utils.py
drhaz: This is a (very nice, by the way) cleanup of my original fitting code. The approach (likelyhood fitting w/ error estimate) Is great, maybe a bit overkill. But it give you good error estimates, which is definitely a plus.

banzai/photometry.py
```diff
@@ -241,3 +243,40 @@ def do_stage(self, image):
...
+        # Match the catalog to the detected sources
+        matched_catalog = match_catalogs(image.catalog, reference_catalog)
+
+        # catalog_mag = instrumental_mag + zeropoint + color_coefficient * color
+        # Fit the zeropoint and color_coefficient rejecting outliers
+        # Note the zero index here in the filter name is because we only store teh first letter of the filter name
+        zeropoint, zeropoint_error, color_coefficient, color_coefficient_error = \
+            fit_photometry(matched_catalog, image.filter[0], self.color_to_fit[image.filter], image.exptime)
+        # Save the zeropoint, color coefficient and errors to header
+        image.meta['L1ZP'] = zeropoint, "Instrumental zeropoint [mag]"
```
drhaz, line 274: Possibly add "color-corrected" to key word description is character limit allows. 

The zeropoint as presented is for stars with a reference color==0, and only for star-like spectral energy distributions.  The danger is to apply a blanket zeropoint to all sources and think we are done - that is not the case. 

At least the BANZI web page (is thereone? read the docs?) should have a comment to this end.

banzai/photometry.py
cmccully: This is definitely a design choice. The primary reason is so we can estimate useful color terms for non stellar SEDs. As we do not know the user's target of interest and are not generating all sky catalogs, I would argue that we should use closer colors. Open to discussion though.

banzai/tests/test_photometry.py
cmccully: This is correct. Though, this is why we explicitly set the random seed here. This should be "reproducibly random".

banzai/photometry.py
cmccully: My hope is that since there is explicitly a color term in the header that it should not be confusing. We should indeed add more high level description to the docs though.

mgdaily (approved):

PR #319, opened 2021-09-01
Update to include "Photometric standards" PROPID in public proposals list

mgdaily:
This will allow standards taken with the new PROPID to be immediately public upon reduction.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+1.3.5 (2021-08-02)
+------------------
+- Added new public LCO standard observation proposals to settings
+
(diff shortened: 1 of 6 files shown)


nvolgenau (commented):
In the portal, "Photometric standards" is capitalized.

nvolgenau (commented):
The data from the proposal codes "FLOYDS standards" and "NRES standards" should also be publicly available.

nvolgenau (commented):
Never mind. The wildcards around *standard* will cover both capitalization and other-standard-propid cases.

nvolgenau (approved):

PR #346, opened 2022-04-04
Fix/opensearch

mgdaily:
This updates to use OpenSearch.

* Removed the timestamp parameter from the OpenSearch call as it's been deprecated.
* Removed the doc type as well - we were getting an error because all docs in OpenSearch are of type `_doc`, and using the update API with doc type `qc` threw an error.
* Updated the nomenclature everywhere to reference OpenSearch instead of ElasticSearch


banzai/main.py
```diff
@@ -88,12 +88,12 @@ def parse_args(settings, extra_console_arguments=None, parser_description='Proce
...
-                        default='http://elasticsearch.lco.gtn:9200')
-    parser.add_argument('--es-index', dest='elasticsearch_qc_index', default='banzai_qc',
-                        help='ElasticSearch index to use for QC results')
-    parser.add_argument('--es-doc-type', dest='elasticsearch_doc_type', default='qc',
-                        help='Elasticsearch document type for QC records')
+    parser.add_argument('--opensearch-url', dest='opensearch_url',
+                        default='https://opensearch.lco.global/')
+    parser.add_argument('--os-index', dest='opensearch_qc_index', default='banzai_qc',
+                        help='OpenSearch index to use for QC results')
+    parser.add_argument('--os-doc-type', dest='opensearch_doc_type', default='qc',
```
mgdaily, line 95: I guess we can remove this since it's not being used explicitly anymore.

helm-chart/banzai/templates/_helpers.tpl
```diff
@@ -84,8 +84,8 @@ External service URLS
   value: {{ .Values.banzai.configdbUrl | quote }}
 - name: OBSERVATION_PORTAL_URL
   value: {{ .Values.banzai.observationPortalUrl | quote }}
-- name: ES_URL
-  value: {{ .Values.banzai.elasticSearchURL | quote }}
+- name: OS_URL
```
jashan-lco, line 87: Can we stick to `OPENSEARCH_URL` for consistency across all projects?

jashan-lco (approved):
Looks good. Don't see any obvious errors. Just a minor nit-pick with `OPENSEARCH_URL` env name.

helm-chart/banzai/templates/_helpers.tpl
mgdaily: Yessir. Good call

banzai/utils/qc.py
```diff
@@ -42,17 +42,16 @@ def save_qc_results(runtime_context, qc_results, image, **kwargs):
     File name, site, camera, dayobs and timestamp are always saved in the database.
     """
 
-    es_output = {}
-    if getattr(runtime_context, 'post_to_elasticsearch', False):
+    os_output = {}
```
cmccully, line 45: os is too overloaded here. We need to say opensearch or something.

banzai/utils/qc.py
```diff
@@ -42,17 +42,16 @@ def save_qc_results(runtime_context, qc_results, image, **kwargs):
...
+    if getattr(runtime_context, 'post_to_opensearch', False):
         filename, results_to_save = format_qc_results(qc_results, image)
-        es = elasticsearch.Elasticsearch(runtime_context.elasticsearch_url)
+        os = OpenSearch(runtime_context.opensearch_url, timeout=60)
```
cmccully, line 48: Same comment here. os is a native python package.

cmccully (approved):

PR #347, opened 2022-04-08
add_bpm function updated to add_cal function

PK0207:
Updated to add_bpm function to reflect the fact that the function can take any kind of calibration file: dark, bias, sky flat, and bad pixel mask.

- Changed the function name to add_cal, and updated the command line function to read in the function.
- The command line function is now `banzai_add_cal`
- Updated the test_end_to_end.py script to use the new function name
- Updated the command line descriptions and error messages to apply to general calibration files rather than just BPM files

Tested the change using the development archive and a local database, and using a bias, dark and skyflat file. All three uploads were successful, as reflected by the database and the archive:
**Database**
image
**Archive**
http://archive-api-dev.lco.gtn/frames/

### DIFF
--- a/banzai/main.py
+++ b/banzai/main.py
@@ -301,22 +301,22 @@ def add_cal():
     logs.set_log_level(args.log_level)
     frame_factory = import_utils.import_attribute(settings.FRAME_FACTORY)()
     try:
-        bpm_image = frame_factory.open({'path': args.filepath}, args)
+        cal_image = frame_factory.open({'path': args.filepath}, args)
     except Exception:
         logger.error(f"Calibration file not able to be opened by BANZAI. Aborting... {logs.format_exception()}", extra_tags={'filename': args.filepath})
-        bpm_image = None
+        cal_image = None
 
     # upload BPM via ingester
     if bpm_image is not None:
         with open(args.filepath, 'rb') as f:
             logger.debug("Posting Calibration file to s3 archive")
-            ingester_response = file_utils.post_to_ingester(f, bpm_image, args.filepath)
+            ingester_response = file_utils.post_to_ingester(f, cal_image, args.filepath)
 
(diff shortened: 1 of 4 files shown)


banzai/main.py
```diff
@@ -288,9 +288,9 @@ def update_db():
         logger.error('Could not populate instruments table: {error}'.format(error=logs.format_exception()))
 
 
-def add_bpm():
-    parser = argparse.ArgumentParser(description="Add a bad pixel mask to the db.")
-    parser.add_argument('filepath', help='Full path to Bad Pixel Mask file')
+def add_cal():
```
mgdaily, line 291: Nitpick, but we could call this `add_super_calibration` - since that a bit more descriptive about what it actually does.

We're still working on updating our nomenclature in the pipeline to get rid of the term "master," so "super" is our new replacement.

banzai/main.py
```diff
@@ -288,9 +288,9 @@ def update_db():
...
+    parser = argparse.ArgumentParser(description="Add a calibration file (bad pixel mask, bias, dark, skyflat) to the db.")
```
mgdaily, line 292: "Add a super calibration file to the db." maybe?

I generally try to avoid hard-coding lists of of the type of calibrations since they can get stale if we were to add a new one and not update it.

mgdaily (approved):
Nice! Looks great! Just a few nitpicky comments about naming things, mostly.

setup.cfg
```diff
@@ -154,7 +154,7 @@ console_scripts =
     banzai_migrate_db = banzai.utils.db_migration:migrate_db
     banzai_add_instrument = banzai.main:add_instrument
     banzai_add_site = banzai.main:add_site
-    banzai_add_bpm = banzai.main:add_bpm
+    banzai_add__super_calibration = banzai.main:add_super_calibration
```
mgdaily, line 157: Looks like there's a double-underscore here between 'add' and 'super'

PR #349, opened 2022-08-22
Reject flat field frames with too low of signal to noise.

cmccully:


banzai/flats.py
```diff
@@ -6,10 +6,26 @@
...
 logger = logging.getLogger('banzai')
 
 
+class FlatSNRChecker(Stage):
+    def __init__(self, runtime_context):
+        super(FlatSNRChecker, self).__init__(runtime_context)
+
+    def do_stage(self, image):
+        # Make sure the median signal-to-noise ratio is over 50 for the image other abort
+        flat_snr = np.median(image.data.signal_to_noise())
```
mgdaily, line 20: I think we can actually do `image.primary_hdu.signal_to_noise()` - `image.data` just returns `image.primary_hdu.data`, which is a numpy array.

Maybe we can write a quick test for this one?

mgdaily (approved):

PR #352, opened 2022-12-02
Simplify Processed Path for Steward Images

griffin-h:
This removes the directory structure for saving Steward images, so that the four 90prime chips are not stored separately. Now the data goes directly to `processed_path/dayobs/` directories.

To do this, I had to define `dayobs` for Steward images. I adopted a convention of using the previous local date if images are taken before local noon


griffin-h:
Sorry, I opened this on the wrong fork!

PR #353, opened 2022-12-02
Distortion correction

griffin-h:
This allows for BANZAI to request that the Gaia astrometry.net service apply a fixed distortion (SIP) correction by sending the camera name and copying the SIP coefficients from the response into the header.


griffin-h:
Related to https://github.com/LCOGT/gaia-astrometry.net-service/pull/23

PR #354, opened 2022-12-19
Update helm chart version to 1.9.9

mgdaily:
This was woefully out-of-date for some reason.


mgdaily:
Odd. It appears the runner can't find tox even though we explicitly install it in one of the earlier steps. I'll look into this a little later, but this corrects the chart @jashan-lco

jashan-lco (approved):

PR #374, opened 2023-10-05
Migrate to no longer use SEP and use astropy Photutils instead.

cmccully:


banzai/photometry.py
```diff
@@ -23,95 +26,125 @@ def radius_of_contour(contour, source):
...
+
+            sources['flag'] = 0
+
+            # Flag = 1 for sources with bad pixels
+            flag_sources(sources, catalog.labels, deblended_seg_map, image.mask, flag=1, mask_value=1)
+            # Flag = 2 for sources that are deblended
+            flag_deblended(sources, catalog, segmentation_map, deblended_seg_map, flag_value=2)
+            # Flag = 4 for sources that have saturated pixels
+            flag_sources(sources, catalog.labels, deblended_seg_map, image.mask, flag=4, mask_value=2)
+            # Flag = 8 if kron aperture falls off the image
```
jchate6, line 142: should `flag_edge_sources` be called here?

banzai/photometry.py
```diff
@@ -23,95 +26,125 @@ def radius_of_contour(contour, source):
...
+    sources_off = np.logical_or(sources_off, minor_ymax > ny)
+    sources_off = np.logical_or(sources_off, major_ymax > ny)
+    sources[sources_off] |= flag_value
 
 
 class SourceDetector(Stage):
-    # Note that threshold is number of sigma, not an absolute number because we provide the error
-    # array to SEP.
-    threshold = 10.0
+    threshold = 2.5
```
jchate6, line 71: where did this threshold come from? Is it no longer a number of sigma but an absolute number?

banzai/photometry.py
```diff
@@ -23,95 +26,125 @@ def radius_of_contour(contour, source):
...
-            sources = Table(sources)
-
-            # We remove anything with a detection flag >= 8
-            # This includes memory overflows and objects that are too close the edge
-            sources = sources[sources['flag'] < 8]
+            segmentation_map = detect_sources(convolved_data, self.threshold, npixels=self.min_area)
+
+            logger.info('Deblending sources', image=image)
+            # Note that nlevels here is DEBLEND_NTHRESH in source extractor which is 32 by default
+            deblended_seg_map = deblend_sources(convolved_data, segmentation_map,
```
jchate6, line 107: this is the step that kills the memory?

jchate6 (changes requested):
A few comments, but nothing major stands out.

banzai/photometry.py
cmccully: Yes

banzai/photometry.py
cmccully: Good catch

jchate6 (approved):

PR #377, opened 2023-10-31
Some environment updates and adding docker builds to ghcr

cmccully:


banzai/tests/e2e-k8s.yaml
```diff
@@ -80,7 +79,7 @@ spec:
         periodSeconds: 1
         timeoutSeconds: 10
     - name: banzai-celery-workers
-      image: @BANZAI_IMAGE@
+      image: banzai:test-latest
```
sfoale, line 82: is this change intended? It looks like you're using a locally built image.

cmccully: It is. This is a longer term fix we are making as we move away from jenkins.

PR #391, opened 2024-04-23
Photometry omit large sources

cmccully:
This PR removes sources with an area of more than 1000 pixels before we attempt to deblend or calculate flux radii. This should prevent long-running photometry issues like we had on tfn1m014-fa20-20240306-0175-e91. 

Matt I recommend running your photometry gist to confirm: https://gist.github.com/mgdaily/aabb6fc8d97a50a09ec15315ce4aab63


mgdaily (approved):
Can confirm this takes 7 seconds on the formerly problematic image. fluxfrac_radius now runs to completion almost instantaneously. Nice work

PR #396, opened 2024-10-04
Handle the base exception from the requests library

mgdaily:
We were handling one of the subclasses, and some errors slipped through, causing missed reductions.

Also update our deploy to use in-cluster URLs to speed up the DNS resolution and make us more resilient to DNS issues.

Sorry for all of the whitespace changes, but it looks like we had some extra end-of-line whitespace in the changelog.

### DIFF
--- a/helm-chart/banzai/values-prod.yaml
+++ b/helm-chart/banzai/values-prod.yaml
@@ -6,20 +6,20 @@
 # will be started when the CPU usage rises above the configured threshold.
 horizontalPodAutoscaler:
   enabled: true
-  minReplicas: 15
-  maxReplicas: 15
+  minReplicas: 25
+  maxReplicas: 25
   targetCPUUtilizationPercentage: 50
-  minLargeReplicas: 7
-  maxLargeReplicas: 7
+  minLargeReplicas: 12
+  maxLargeReplicas: 12
 
 image:
   repository: ghcr.io/lcogt/banzai
-  tag: "1.18.1"
+  tag: "1.18.2"
   pullPolicy: IfNotPresent
 
 # Values for the OCS Ingester library, used by BANZAI.
 ingester:
-  apiRoot: http://archiveapi-internal.prod/
+  apiRoot: http://archiveapi-internal/
   s3Bucket: archive-lco-global
   ingesterProcessName: banzai_imaging
   opentsdbHostname: opentsdb.lco.gtn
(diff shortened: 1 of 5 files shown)


helm-chart/banzai/values-prod.yaml
```diff
@@ -6,20 +6,20 @@
 # will be started when the CPU usage rises above the configured threshold.
 horizontalPodAutoscaler:
   enabled: true
-  minReplicas: 15
-  maxReplicas: 15
+  minReplicas: 25
```
jchate6, line 9: We should probably talk about how to introduce autoscaling someday when you have the bandwidth...

helm-chart/banzai/values-prod.yaml
```diff
@@ -6,20 +6,20 @@
...
   targetCPUUtilizationPercentage: 50
-  minLargeReplicas: 7
-  maxLargeReplicas: 7
+  minLargeReplicas: 12
+  maxLargeReplicas: 12
 
 image:
   repository: ghcr.io/lcogt/banzai
-  tag: "1.18.1"
+  tag: "1.18.2"
```
jchate6, line 17: Should this match the version number above in the changes? That's currently 1.18.3...

jchate6 (approved):
I don't really understand the reason for the banzai changes with urls, but I trust that you do.

I have an actual question about versioning, but again, I might be missing things, so I'm clicking approve.

helm-chart/banzai/values-prod.yaml
cmccully: Do we need to set a max replicas here? Or just assume if we only set a min, they are equal?

mgdaily: > We should probably talk about how to introduce autoscaling someday when you have the bandwidth...

Jashan did prototype this at one time. https://github.com/LCOGT/banzai/pull/359/files

Autoscaling based on queue size _sounds_ great Joey, but I am very hesitant to commit my time to this as a development priority. I think it is relatively high effort to low gain, given that we can simply run more pods than we need most of the time.

I would love if someone other than me could spearhead that effort.

mgdaily: > Do we need to set a max replicas here? Or just assume if we only set a min, they are equal?

Yes we need to. https://github.com/LCOGT/banzai/pull/359/files

This has always been the case.

helm-chart/banzai/values-prod.yaml
mgdaily: Good catch.

helm-chart/banzai/values-prod.yaml
cmccully: Yeah. Ok. github split the file and I missed the next line.

PR #397, opened 2024-10-16
Added the ability to read fits files that are pre downloaded and are already in memory

cmccully:


sfoale (approved):
Looks good. Simple enough for even me to understand.

mgdaily (approved):

PR #404, opened 2025-02-04
Add TSO to list of PUBLIC_PROPOSALS

nvolgenau:
Add TSO to list of PUBLIC_PROPOSALS. TSO is prefix for The Schools Observatory, which assumed management of the Faulkes Telescope Project.


cmccully:
I added a note to the changelog but otherwise easy enough.

cmccully (approved):

PR #405, opened 2025-02-12
Fixes to migration from cals by block.

cmccully:
This breaks the alembic migration into two parts so we shouldn't break the database even if back populating data takes a while. I've also restructured how the archive API calls are made so we don't get rate limited.

In testing I ran into significant issues with old versions of deps. I updated the deps to more modern versions.


jchate6 (commented):
There are a lot of failing style checks and tests etc. Should we try to fix those?
I can help with the style things a least.

sfoale (approved):
looks good to me

PR #407, opened 2025-03-17
Reverting changes on how calibrations are chosen that were included for floyds

cmccully:
I've removed the prefer-same-block-cals, check-public-cals, and prefer-same-proposal-cals options that were being used by floyds as none of the other pipelines need them and it was causing too many dependency issues. For floyds, I will inject these changes by overriding the default CalibrationUser class.


jchate6 (approved):
Do we want to add additional logging anywhere?

PR #408, opened 2025-03-18
Revert "Reverting changes on how calibrations are chosen that were included for floyds"

jchate6:
Reverts LCOGT/banzai#407


jchate6:
I don't understand why e2e tests are failing, but it doesn't feel like forcing this push will be worse than leaving it as it is all night...

PR #412, opened 2025-05-13
run worker with --task-events

jashan-lco:
Needed by Flower


jchate6 (approved):
looks good assuming the e2e tests ever clear.
I'd probably wait for Curtis's feedback before deploying though?

PR #414, opened 2025-06-18
Downgrade astropy to 6 because 7.1 produces files that are not compatible with cfitsio.

cmccully:


talister (approved):
Looks good. We could possibly tighten it to `<7.0.2` but if it was previously on 6.1, then "not 7" is safer. Once bitten, twice shy and all that...

markBowman (approved):
LGTM

PR #415, opened 2025-06-19
Fix example usage of create_db in README

timbeccue:
Change the example usage of `create_db` to use a single argument. 

Previously, it used `create_db('.', db_address='sqlite:///banzai.db')`, but it looks like `create_db` should have the db_address as its only argument.


Fingel:
Ran into this today as well, @timbeccue you should add a reviewer or this won't be able to merge.

cmccully (approved):

PR #416, opened 2025-06-25
Rename banzai.celery -> banzai.schedule

Fingel:
Resolves a naming conflict with the Celery library:

ImportError: cannot import name 'Celery' from 'celery' (consider renaming '/home/austin/Documents/lco/code/banzai/banzai/celery.py' if it has the same name as a library you intended to import)

If/when banzai moves away from Celery this will make more sense naming-wise.


cmccully:
I think this is required by python 3.13 so that's good. I want to get the e2e tests running before I merge because for at least some version of celery there were hardcoded module name expectations within celery.

cmccully (approved):

PR #419, opened 2025-07-01
119 opentelemetry for banzai

Fingel:
Instruments some functions and stages with spans allowing for the
visibility of execution time for the pipeline. There will need to be an
OpenTelemetry compatible collector running to view the traces as
anything other than JSON. This implementation used a Jaeger backend but
they should be generic.

There was some effort put into making sure Banzai would still run even
if opentelemtry was disabled/not installed.

Two new env vars have been added:

`BANZAI_ENABLE_TRACING`: Enable tracing if set to 'true', '1', or 'yes'.'
`OTEL_EXPORTER_OTLP_ENDPOINT`: The endpoint for exporting traces. If
omitted defaults to the console.


Fingel:
image
Example Jaeger trace of an image reduction.

Fingel:
The deployment plan for this should be pretty simple on the Banzai end:

1. If this PR is merged, nothing needs to happen. In fact, if the OpenTelemetry libs aren't installed most of everything in this PR becomes a no-op.
2. If the libs are installed but there is no consumer (Jaeger) set up yet, we can set it to console output. So it will output specialized logs that aren't very useful without a machine to pick them up.
3. Eventually we need Jaeger installed somewhere to actually push the spans. But that doesn't need to necessarily happen synchronously with this PR being merged.

PR #420, opened 2025-07-03
Fix issue preventing use of local BPM files

timbeccue:
This PR modifies the `add_super_calibration` function (exposed to the console as `banzai_add_super_calibration`), adding an optional boolean flag `--skip-ingester` that will take the calibration file and add it directly to the database without posting to the ingester. 

Before making this change, I was unable to use a local BPM file because it was failing the ingester step. It seems like that shouldn't be a required step, nor is it one that most users would have permission to do anyways. 

I also added an optional arg (`--frame-id`) to manually pass in the calibration frame_id, which is otherwise sourced from the ingester step. However, it seems to work just fine without this. Actually, regardless of whether or not this value is supplied, manually querying the db for the bpm file shows a null entry for db `frameid`. Do we need this at all?

### DIFF
--- a/banzai/main.py
+++ b/banzai/main.py
@@ -314,7 +314,7 @@ def add_super_calibration():
     parser.add_argument('--db-address', dest='db_address',
                         default='mysql://cmccully:password@localhost/test',
                         help='Database address: Should be in SQLAlchemy form')
-    parser.add_argument('--skip-ingester', dest='skip_ingester', action='store_true')
+    parser.add_argument('--upload-to-archive', dest='upload_to_archive', action='store_true')
     args = parser.parse_args()
     add_settings_to_context(args, settings)
     logs.set_log_level(args.log_level)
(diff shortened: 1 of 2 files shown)


banzai/main.py
```diff
@@ -314,31 +314,45 @@ def add_super_calibration():
...
         cal_image = frame_factory.open({'path': args.filepath}, args)
     except Exception:
         logger.error(f"Calibration file not able to be opened by BANZAI. Aborting... {logs.format_exception()}",
                      extra_tags={'filename': args.filepath})
-        cal_image = None
+        return
+
+    if args.skip_ingester:
+        logger.debug("Skipped posting frame to archive. Saving to database.", extra_tags={'frameid': args.frame_id})
+        cal_image.frameid = args.frame_id
```
cmccully, line 333: If we aren't going to the ingester, the frame id isn't really meaningful (that is the archive frame id). Things should work if frame id is none (i.e. not in the archive) but the row has a filepath and a filename.

cmccully (approved):

PR #421, opened 2025-07-15
Add a notebook with runnable manual pipeline reduction

timbeccue:
The primary change here is the addition of a notebook that runs through a full reduction in the manual pipeline mode, inspired by its [nres counterpart](https://banzai-nres.readthedocs.io/en/latest/banzai_nres/ExampleReduction.html). 

In addition, there are a few readme bugfixes and changes for added clarity based on my experience as a first time user setting things up from scratch.

This pr depends on [420](https://github.com/LCOGT/banzai/pull/420), which should be merged first.


timbeccue:
Talked with Curtis and Joey about implementing the following changes:
- change the default db address from 'banzai.db' to 'banzai-test.db'
- in the example, include the option to use files directly from the archive using filename and frame_id
- add the notebook to the readthedocs (like in banzai-nres)
- update the version in pyproject.toml and update changes.md

cmccully (approved):

PR #423, opened 2025-07-22
add development workflow w/ skaffold

jashan-lco:
- Make Archive API requests w/ `include_related_frames=false` to avoid timing out when intializing DB locally.
- Add a devenv Flake
- Add base Kustomization and local development overlay for Skaffold

Integration tests run, but some of them are failing with duplicate key constraints. Need to debug that further.

### DIFF
--- a/k8s/base/cronjob-db-bpm-update/cronjob.yaml
+++ b/k8s/base/cronjob-db-bpm-update/cronjob.yaml
@@ -1,0 +1,73 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/master-standalone/cronjob-batch-v1.json
+apiVersion: batch/v1
+kind: CronJob
+metadata: # kpt-merge: /example
+  name: db-bpm-update
+  labels:
+    app.kubernetes.io/component: db-bpm-update
+  annotations:
+    internal.kpt.dev/upstream-identifier: 'batch|CronJob|default|example'
+spec:
+  schedule: "@daily"
+  concurrencyPolicy: Forbid
+  failedJobsHistoryLimit: 1
+  successfulJobsHistoryLimit: 3
+  startingDeadlineSeconds: 120
+  jobTemplate:
+    spec:
+      backoffLimit: 3
+      activeDeadlineSeconds: 3600
+      template:
+        spec:
+          restartPolicy: Never
+          initContainers:
+            # Create the db if it doesn't exist
+            - name: init-db
+              image: "banzai"
+              command:
+                - "banzai_create_db"
+                - "--db-address=$(DB_ADDRESS)"
+              envFrom:
+                - configMapRef:
+                    name: env
+                    optional: false
+                - secretRef:
+                    name: db-env
+                    optional: true
+                - secretRef:
+                    name: rabbitmq-env
+                    optional: true
+                - secretRef:
+                    name: archive-api-env
+                    optional: true
+                - secretRef:
+                    name: env
+                    optional: true
+          containers:
+            - name: default
+              image: "banzai"
+              command:
+                - "banzai_populate_bpms"
+                - "--db-address=$(DB_ADDRESS)"
+              envFrom:
+                - configMapRef:
+                    name: env
+                    optional: false
+                - secretRef:
+                    name: db-env
+                    optional: true
+                - secretRef:
+                    name: rabbitmq-env
+                    optional: true
+                - secretRef:
+                    name: archive-api-env
+                    optional: true
+                - secretRef:
+                    name: env
+                    optional: true
+        metadata:
+          labels:
+            app.kubernetes.io/component: db-bpm-update
+    metadata:
+      labels:
+        app.kubernetes.io/component: db-bpm-update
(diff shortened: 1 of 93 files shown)


Fingel (approved):
Mark asked me to review this so I'm going to insert myself into the reviewers.

I managed to follow the instructions in DEVELOPMENT_K8s.md in a fresh virtual machine and got everything up and running, at least until I got to the point where I could run the skaffold e2e commands.

A few notes:

* Instead of `skaffold -m banzai run` I had more success with `skaffold -m banzai debug/dev`
* Make sure you `skaffold delete` before re-running and debug/dev commands for banzai or it's deps, otherwise skaffold appears to try and deploy new copies that interfere with the old ones.
* `skaffold debug`  seems to fail with some commands because of a missing python dependency, pydbg or something along those lines, lost to scrollback at this point.

The e2e are indeed failing. Which is not surprising to me as it took me a long time to get them to pass locally, and it's still inconsistent for me.

One thing that was a issue for me was adding the initial bad pixel masks (BPM files) for the cameras I wanted to test. I had to do it manually and that that got me further along towards passing e2e tests. @LTDakin can chime in here. Unfortunately my unfamiliarity with the K* tools means I'm not sure how I would do the same thing in this environment. But this might be worth looking into.

All in all this seems like some impressive stuff. However due to my own unfamiliarity with all the layers of tooling it will probably be a while before I use this for any kind local development.

jashan-lco:
>     * Instead of `skaffold -m banzai run` I had more success with `skaffold -m banzai debug/dev`

That's interesting. What was the error with `run`? `dev` is using the same pipeline under the hood, but in "watch" mode, so it's strange that it works but `run` does not.
> 
>     * Make sure you `skaffold delete` before re-running and debug/dev commands for banzai or it's deps, otherwise skaffold appears to try and deploy new copies that interfere with the old ones.

AFAIK, `run/dev/debug` should clobber over parts that need clobbering (like updating the docker image) instead of deploying duplicates.

> 
>     * `skaffold debug`  seems to fail with some commands because of a missing python dependency, pydbg or something along those lines, lost to scrollback at this point.

Yes, `debug` is intended to be used with a debugger, so requires extra dependencies. One way to do this (in the future) might be to add an optional dependency group to pyproject.toml and build a "debug" container with that installed when the command is `skaffold debug`, otherwise default to the standard dependencies.

> 
> The e2e are indeed failing. Which is not surprising to me as it took me a long time to get them to pass locally, and it's still inconsistent for me.
> 
> One thing that was a issue for me was adding the initial bad pixel masks (BPM files) for the cameras I wanted to test. I had to do it manually and that that got me further along towards passing e2e tests. @LTDakin can chime in here. Unfortunately my unfamiliarity with the K* tools means I'm not sure how I would do the same thing in this environment. But this might be worth looking into.

Ok, thanks for the tip. I'll look into that.

jashan-lco:
> To add some clarity: it's not that `run` wasn't working more that dev shows more logs and seemed more interactive, so it's easier to tell what's going on.

I see. Yes, by default `run` will just do things in the background. If you don't want `dev`'s hot-reload/watch behavior you can run `run` with `--tail`, e.g. `skaffold -m module run --tail`

> > AFAIK, run/dev/debug should clobber over parts that need clobbering (like updating the docker image) instead of deploying duplicates.
> 
> I'm not sure how but I got into a situation where services were not starting up and throwing errors about being unable to bind to in-use addresses.

The other thing `dev` does is it tries to port-forward *every* `Service` port from your localhost to the docker/kubernetes environment. My guess is one of the Services tries to bind to 80/443. But these ports are already forwarded by the Kuberentes cluster to ingress HTTP traffic into it. (side note: this is how you can access *.local.lco.earth domains from your browser and have them automatically route to things running in the *local* Kubernetes cluster, e.g. https://k8s.local.lco.earth)

flake.lock
```diff
@@ -0,0 +1,587 @@
...
+        "gomod2nix": "gomod2nix",
+        "nixpkgs": [
+          "devenv-k8s",
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1711736775,
+        "narHash": "sha256-wRzy040KH+IfgK+g3Dj4Re4hb1S+G3xaxJ41TrsPEvE=",
+        "owner": "jashandeep-sohi",
```
cmccully, line 543: Why do we use your personal iteration of skaffold?

k8s/base/cronjob-db-bpm-update/cronjob.yaml
```diff
@@ -0,0 +1,73 @@
...
+apiVersion: batch/v1
+kind: CronJob
+metadata: # kpt-merge: /example
+  name: db-bpm-update
+  labels:
+    app.kubernetes.io/component: db-bpm-update
+  annotations:
+    internal.kpt.dev/upstream-identifier: 'batch|CronJob|default|example'
+spec:
+  schedule: "@daily"
```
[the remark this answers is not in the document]
cmccully: Basically, I'm not sure we really want the local dev copy to run the cron tab regularly?

flake.lock
cmccully: I think it would make me feel marginally better to put the fixes onto an LCOGT org fork.

PR #436, opened 2026-01-27
Site deployment - calibration caching

timbeccue:
This PR implements the calibration file caching system for site deployments. This allows sites to maintain local
copies of calibration files, enabling image processing during network outages.

```mermaid
sequenceDiagram
      participant AWS as AWS PostgreSQL
      participant Local as Site PostgreSQL
      participant Worker as Download Worker
      participant Archive as Archive API
      participant Cache as Cache Directory

      AWS->>Local: Replicate calibration metadata
      Note over Local: filepath = NULL (from AWS)

      loop Every poll interval
          Worker->>Local: Query needed calibrations (top 2 per config)
          Local-->>Worker: Return calibrations without local files
          Worker->>Archive: Download calibration file
          Archive-->>Worker: FITS file data
          Worker->>Cache: Write file to disk
          Worker->>Local: UPDATE db filepath = '/data/cal/file.fits'

          Worker->>Cache: Delete old versions (keep top 2)
          Worker->>Local: Clear db filepath for deleted files
      end

      AWS->>Local: Replication update (filepath=NULL)
      Note over Local: Trigger preserves<br/>local db filepath
```

  ## Key components:
  - PostgreSQL logical replication setup to sync calibration metadata from AWS to local DB
  - Download worker daemon that polls the local DB and fetches calibration files from the archive
  - SQL trigger to preserve local file paths during replication
  - Version management (keeps top 2 calibrations per instrument/type/config)

## Changes

  - banzai/cache/ - New package with replication setup, download worker, and init container logic
  - banzai/dbs.py - Added CacheConfig model and CachedCalibrationInfo dataclass
  - sql/triggers.sql - Trigger to prevent replication from overwriting local filepaths
  - docker-compose-site.yml - Updated for cache system integration

## Tests

Site e2e tests are separate from the main e2e tests because it involves a different configuration.
For the site e2e tests, we mock the AWS banzai db with a local postgres instance that includes publication.  We also run the docker-compose-site.yml setup and subscribe to the local publication. Then we populate the publication db with some calibration frames and test:
- replication to our local db works
- the latest two versions of each calibration are cached
- the local worker is able to reduce a raw frame using these cached calibrations
- adding new calibrations removes any now-stale cached files.

Unit tests are also included for the files in `banzai/cache/`.

Note: this PR is based on banzai-local and should be merged after that branch lands


banzai/cache/download_worker.py
```diff
@@ -0,0 +1,191 @@
...
+                dbs.CalibrationImage.frameid, dbs.CalibrationImage.type,
+                dbs.CalibrationImage.dateobs, dbs.CalibrationImage.filepath,
+                dbs.Instrument.site.label('site'), dbs.Instrument.camera.label('camera'),
+                rank,
+            ).join(dbs.Instrument).filter(
+                dbs.CalibrationImage.is_master == True,
+                dbs.CalibrationImage.is_bad == False,
+                dbs.Instrument.site == self.site_id,
+            )
+            if self.instrument_types != ['*']:
```
cmccully: Will the star get guider frames too? It will do a lot of extra null queries but no real harm I guess.

banzai/cache/download_worker.py
```diff
@@ -0,0 +1,191 @@
...
+        self.db_address = db_address
+        self.site_id = site_id
+        self.instrument_types = instrument_types
+        self.processed_path = processed_path
+        self.runtime_context = runtime_context
+
+    def get_calibrations_to_cache(self):
+        """Return top 2 calibrations per config via SQL window function."""
+        with dbs.get_session(self.db_address) as session:
+            config_mode = cast(dbs.CalibrationImage.attributes['configuration_mode'], String)
```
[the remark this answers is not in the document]
timbeccue: yeah they're json attributes, and we need them as strings to work in `partition_by`

banzai/cache/download_worker.py
timbeccue: This is a configurable value that can be modified with the `INSTRUMENT_TYPES` variable. The star is a default if nothing is provided.

cmccully (approved):

PR #444, opened 2026-03-18
Smartstack Scaffolding

timbeccue:
Originally this PR was just splitting Redis/RabbitMQ out of the site/local
docker-compose files, but it ended up bundling several dependent changes
together rather than stacking PRs that kept overriding each other.

What's in here now:

### Smart stacking scaffolding
- New `StackFrame` model + `SiteBase` declarative base for site-only tables
  (`dbs.py`)
- `SubframeListener` + `process_subframe` Celery task: validates incoming
  RabbitMQ messages, inserts a tracking row, runs the reduction pipeline,
  updates the row with the reduced filepath, and pushes a Redis notification
  per camera (`main.py`, `scheduling.py`)
- `stacking.py`: per-camera worker loop drains Redis notifications, checks
  group completion (all reduced AND all expected frames present, or
  `is_last`), and finalizes complete or timed-out stacks; supervisor spawns
  one worker per camera
- New `banzai-subframe-listener`, `banzai-subframe-worker`,
  `banzai-stacking-supervisor` services + matching env vars
- E2E coverage in `test_smart_stacking.py` and a new site-E2E phase

### Redis / RabbitMQ as external dependencies
- Pulled out of `docker-compose-{site,local}.yml` into
  `docker-compose-dependencies.yml` so they can be left running across
  pipeline restarts (and pointed at existing site infrastructure)
- All workers connect via `REDIS_URL` / `RABBITMQ_URL`; no more hardcoded
  `redis://redis:6379/0`

### Site deployment fixes & polish (from running this at a site)
- Download worker: clearer status logging w/ heartbeat + state-change
  triggers, NULL-`frameid` handling, per-frameid failure backoff, filepath
  reconciliation when replication overwrites with NULL, "waiting for
  replication" log while `calimages` is empty during initialization
- Replication: fixed subscription name, optional `SLOT_NAME` override,
  reuse existing slot when recreating subscription
- Listener improvements: log and remove malformed JSON payloads from queue, top-level try/except around the stacking worker loop
- `created_at` (not `dateobs`) for stack-timeout — handles NULL `dateobs`
  and reprocessing (note: stack timeout is being reworked a bit in the branch I'm currently working on)
- Site E2E: Linux compatibility (named volume for postgres,
  `host.docker.internal` resolution), container-prefix support so the
  fixture stack and `just site-up` can coexist, absolute-`HOST_*_DIR`
  requirement so file paths resolve identically inside and outside
  containers
- `argparse` entry points for site workers; removed unused
  `OPENTSDB_PYTHON_METRICS_TEST_MODE` plumbing


banzai/cache/download_worker.py
```diff
@@ -12,7 +13,8 @@
 from banzai.utils import date_utils, file_utils, fits_utils
 
 logger = logs.get_logger()
-HEARTBEAT_INTERVAL = 300
+HEARTBEAT_INTERVAL = 600        # seconds
+FAILURE_RETRY_SECONDS = 43200   # seconds
```
cmccully: That's a really long time in seconds... Do we really want to wait this long?

banzai/cache/download_worker.py
```diff
@@ -92,7 +108,8 @@ def download_calibration(db_address, processed_path, runtime_context, cal):
     finally:
         buffer.close()
     update_filepath(db_address, cal.id, dest_dir)
-    logger.info(f"Downloaded {cal.filename}")
+    logger.info(f"Cached {cal.filename}")
+    return True
```
cmccully: I don't generally like returning True unless it is a boolean check, e.g., is this object x type? I tend to raise an exception if something didn't work. This is kind of a stylistic thing, but it's a place to make the code more readable for future us.

banzai/cache/download_worker.py
```diff
@@ -121,67 +149,130 @@ def run_download_worker(db_address, site_id, instrument_types, processed_path,
...
+                    f"{deleted_count} deleted"
+                )
 
             time.sleep(poll_interval)
         except Exception as e:
             logger.error(f"Error in worker loop: {e}", exc_info=True)
             time.sleep(30)
 
 
+def create_parser():
```
cmccully: I'm on the fence about whether I like this logic being a function. If it's reused anywhere, it should definitely be a function. It does help the readability of the other function a little at the cost of splitting the logic into multiple places. Maybe the function just needs a more descriptive name to signal to the reader that this is specific to this module and not some generic library function.

banzai/cache/init.py
```diff
@@ -35,16 +35,16 @@ def run_initialization():
 
     try:
         logger.info("Creating database schema...")
-        dbs.create_db(args.db_address)
+        dbs.create_db(args.db_address, site=True)
```
cmccully: I'm not sure I like this variable being just "site" here. We use site a lot of places and it is a 3 letter site code generally. Maybe at_site? Or site_deploy?

banzai/cache/replication.py
```diff
@@ -1,50 +1,60 @@
...
-    if slot_name is None:
-        slot_name = f'banzai_{site_id}_slot'
 
-    logger.info("Setting up replication subscription",
-                extra_tags={'subscription': subscription_name, 'publication': publication_name,
-                            'slot': slot_name, 'site': site_id})
-
-    subscription_sql = f"""
-    CREATE SUBSCRIPTION {subscription_name}
+def _subscription_sql(aws_connection_string, publication_name, slot_name, create_slot=True):
```
cmccully, line 16: What is a slot in this context? Is this postgres specific? If it's standard nomenclature, then it's fine.

banzai/tests/site_e2e/conftest.py
```diff
@@ -33,20 +36,38 @@ def load_env_file(env_path):
 
 load_env_file(SITE_E2E_DIR / "site_e2e.env")
 
+# HOST_*_DIR paths must be absolute so that host-path:host-path volume
+# mounts in docker-compose-site.yml resolve correctly (mirrors the
+# production site-banzai-env contract).
+for _key in ('HOST_RAW_DIR', 'HOST_CALS_DIR', 'HOST_REDUCED_DIR'):
```
cmccully: Is there any way we can add sensible defaults for these values so that it doesn't just crash out if they are not set?

banzai/tests/site_e2e/conftest.py
```diff
@@ -33,20 +36,38 @@ def load_env_file(env_path):
...
+
 # Configuration constants (from env vars with defaults)
 PUBLICATION_DB_ADDRESS = os.environ.get(
     "PUBLICATION_DB_ADDRESS",
     "postgresql+psycopg://banzai:banzai_test@localhost:5433/banzai_test"
 )
 LOCAL_DB_ADDRESS = os.environ.get(
     "LOCAL_DB_ADDRESS",
-    "postgresql+psycopg://banzai@localhost:5442/banzai_local"
+    f"postgresql+psycopg://banzai@localhost:{os.environ.get('SITE_PG_HOST_PORT', '5443')}/banzai_local"
```
cmccully, line 64: Do we want to make the banzai site db hostname configurable?

banzai/tests/site_e2e/conftest.py
```diff
@@ -65,26 +86,67 @@ def poll_until(predicate, timeout, interval=5):
     return result
 
 
-def run_docker_compose(compose_file, *args, cwd=None, env=None):
+def publish_to_queue(queue_name, body, broker_url='amqp://localhost:5672'):
```
cmccully: What's the difference between this code and the code we use to post to the fits exchange? I guess maybe the fits exchange? This is not an important thing if it is different enough, but just smelled like code duplication.

banzai/tests/site_e2e/conftest.py
```diff
@@ -65,26 +86,67 @@ def poll_until(predicate, timeout, interval=5):
...
+        ["docker", "run", "--rm", "-v", f"{path}:/target", "alpine",
+         "sh", "-c", "rm -rf /target/*"],
+        capture_output=True, timeout=30,
+    )
+    if result.returncode != 0:
+        logger.warning(f"Docker cleanup failed for {path}: {result.stderr}")
+    shutil.rmtree(path, ignore_errors=True)
+
+
+def run_docker_compose(compose_file, *args, cwd=None, env=None, project=None):
```
cmccully, line 191: This may just be me, but I'm really nervous about python calling docker when this python code will likely be running in a docker container.

banzai/tests/site_e2e/README.md
```diff
@@ -68,14 +71,34 @@ pytest -m e2e_site_reduction banzai/tests/site_e2e/ -v
...
+While tests run, you can monitor service logs in separate terminals. The e2e
+fixture brings the site stack up under compose project `banzai-e2e` with the
+container prefix `e2e-banzai-`:
 
 ```bash
-docker logs -f banzai-download-worker  # Watch file downloads
-docker logs -f banzai-cache-init       # Watch replication setup
-docker logs -f banzai-worker           # Watch frame processing
+docker logs -f e2e-banzai-download-worker  # Watch file downloads
+docker logs -f e2e-banzai-cache-init       # Watch replication setup
```
cmccully, line 138: My github review panel is showing these lines as grey. Just want to double check they render correctly outside of the diff.

banzai/tests/site_e2e/README.md
```diff
@@ -68,14 +71,34 @@ pytest -m e2e_site_reduction banzai/tests/site_e2e/ -v
...
+machine:
+
+- Container names don't collide (`e2e-banzai-worker` vs `banzai-worker`).
+- Queues don't collide (`e2e_reduction_task_queue` vs `reduction_task_queue`),
+  even though both stacks share the same `banzai-redis` / `banzai-rabbitmq`.
+- Postgres host ports don't collide (5443 for e2e, 5442 for site-up).
+- `docker compose down` from the e2e fixture is project-scoped — it cannot
+  tear down a docker-compose-site stack (project `banzai`).
+
+The only manual collision risk is the host data directories (`HOST_RAW_DIR`,
```
cmccully, line 159: This is another place sensible defaults could help.

banzai/tests/site_e2e/site_e2e.env.template
```diff
@@ -23,6 +39,10 @@ AUTH_TOKEN=
 # Download worker
 DOWNLOAD_WORKER_POLL_INTERVAL=1
 
+# Message brokers
+REDIS_URL=redis://host.docker.internal:6379/0
```
cmccully, line 46: I need to go look at the docker compose more carefully. You might want to actually create a docker network here that isn't just the default. The way docker does its DNS is different between Mac and linux so it can bite you.

banzai/tests/site_e2e/test_site_e2e.py
```diff
@@ -148,11 +156,10 @@ def test_06_queue_raw_frame(self, site_deployment, auth_token):
 
         result = subprocess.run(
             [
-                'uv', 'run', 'python', str(queue_script),
+                sys.executable, str(queue_script),
```
cmccully: The subprocesses are getting nested deeply here. Maybe we should think a little more about this deploy.

banzai/tests/site_e2e/test_site_e2e.py
```diff
@@ -205,8 +211,73 @@ def test_07_reduction_completes(self, site_deployment):
...
+        def check():
+            with dbs.get_session(LOCAL_DB_ADDRESS) as session:
+                cal = session.query(dbs.CalibrationImage).filter_by(filename=target).first()
+                return cal.filepath if cal and cal.filepath == expected else None
+
+        assert poll_until(check, timeout=60, interval=2) == expected, (
+            f"worker did not reconcile {target} back to {expected} after drift to {drifted_value!r}"
+        )
+
+    @pytest.mark.e2e_site_cache
```
cmccully, line 279: I don't remember if there is, but is there a way to set a pytest run to skip by default? Running the unit tests not has -m "not e2e" -m "not e2e_site_cache" which is starting to get onerous on a newbie user just trying to spin up the project.

banzai/tests/test_smart_stacking.py
```diff
@@ -0,0 +1,520 @@
...
+        for f in frames:
+            assert f.status == 'timeout'
+            assert f.completed_at is not None
+
+
+# ---------------------------------------------------------------------------
+# Timeout
+# ---------------------------------------------------------------------------
+
+class TestTimeout:
```
cmccully: I would check the naming conventions in pytest. Something generic like this may be a little hard to find if it doesn't print the module name right next to it.

banzai/dbs.py
```diff
@@ -115,6 +128,24 @@ class ProcessedImage(Base):
     tries = Column(Integer, default=0)
 
 
+class StackFrame(SiteBase):
```
cmccully: I'm not sure I like the name of the table or class. frames_to_stack might be epsilon better.

banzai/main.py
```diff
@@ -228,6 +230,63 @@ def start_listener(runtime_context):
             logger.info('Shutting down pipeline listener.')
 
 
+class SubframeListener(ConsumerMixin):
```
cmccully: This is probably a nit, but is there a way to refactor things to reuse more of the regular listener. I'm just trying to avoid code duplication where possible.

banzai/scheduling.py
```diff
@@ -247,3 +250,58 @@ def process_image(self, file_info: dict, runtime_context: dict):
...
+        header = fits_utils.get_primary_header(filepath)
+        if header is None:
+            raise self.retry(countdown=30, exc=IOError(f"Could not read header from {filepath}"))
+
+        camera = header.get('INSTRUME', '').strip()
+        dateobs_str = header.get('DATE-OBS', '')
+        dateobs = parse_date_obs(dateobs_str) if dateobs_str else None
+
+        # Phase 1: Insert DB record before reduction so stacking worker can see it
+        insert_stack_frame(
```
cmccully: Question about the stacking infrastructure. Is the stacking worker checking a database or listening to a queue on say rabbitmq? It seems like a listener would be a better pattern as polling a db for changes is error prone.

banzai/stacking.py
```diff
@@ -0,0 +1,201 @@
...
+def drain_notifications(redis_client, camera):
+    """Drain and return a deduplicated set of moluids from the Redis list for a camera."""
+    key = f'{REDIS_KEY_PREFIX}{camera}'
+    drain_key = f'{key}:draining'
+    # Atomic rename so notifications pushed between read and delete aren't lost
+    try:
+        redis_client.rename(key, drain_key)
+    except redis_lib.exceptions.ResponseError:
+        return set()
+    raw = redis_client.lrange(drain_key, 0, -1)
```
cmccully: This scares me. Can you say why we are having to drain queues manually?

banzai/stacking.py
```diff
@@ -0,0 +1,201 @@
...
+            continue
+        frmtotal = frames[0].frmtotal
+        if check_stack_complete(frames, frmtotal):
+            finalize_stack(db_address, moluid, status='complete')
+
+
+def finalize_stack(db_address, moluid, status='complete'):
+    """Mark stack complete and log mock stacking/JPEG/ingester operations."""
+    dbs.mark_stack_complete(db_address, moluid, status=status)
+    logger.info(f'Mock stacking complete for {moluid}', extra_tags={'moluid': moluid})
```
cmccully: I understand these are mock warnings, but it opens a general point. Info messages should be logs we want to store in something like opensearch that can be used later. The temptation is to leave in debug messages when writing the code. If we can lower those to logger.debug that's fine. banzai is already really noisy and I'm hesitant to add more noise.

banzai/stacking.py
```diff
@@ -0,0 +1,201 @@
...
+            proc = multiprocessing.Process(
+                target=run_worker_loop,
+                args=self._worker_args(camera),
+                name=f'stacking-worker-{camera}',
+            )
+            proc.start()
+            self.workers[camera] = proc
+            logger.info(f'Started stacking worker for camera {camera}')
+
+    def monitor(self, check_interval=10):
```
cmccully, line 153: If feels strange to have the monitoring function and worker function in the same Python class. I'm struggling to visualize the whole infrastructure of processes. My sense is it kind of feels like overkill. Maybe we could a process flow diagram to the docs?

banzai/stacking.py
```diff
@@ -0,0 +1,201 @@
...
+    def shutdown(self):
+        """Graceful shutdown of all workers."""
+        for camera, proc in self.workers.items():
+            proc.terminate()
+            proc.join(timeout=10)
+            logger.info(f'Stopped stacking worker for camera {camera}')
+        self.workers.clear()
+
+
+def create_parser():
```
cmccully: Another create_parser method. I'd try to use a little more descriptive name.

cmccully (changes requested):
I think this is a huge step forward. I've left some comments so more significant than others. Once those are addressed, I'm happy to approve.

banzai/cache/download_worker.py
timbeccue: Probably a poorly chosen balance when I was trying to reduce log clutter from repeated retries. Maybe 2 hours is better than 12? I can improve the readability here too.

banzai/tests/site_e2e/conftest.py
timbeccue: It's currently configurable by setting the value of LCOAL_DB_ADDRESS; localhost is just part of the default value. So I think this is fine unless I'm not understanding your question.

banzai/cache/replication.py
timbeccue: "Slot" is standard postgres terminology. A replication slot is the named server-side object that tracks which WAL records a subscriber has consumed. See https://www.postgresql.org/docs/current/warm-standby.html#STREAMING-REPLICATION-SLOTS.

banzai/cache/download_worker.py
timbeccue: This function takes a file that needs to be cached and results in three possible outcomes: downloaded, skipped, or failed. The return value is used by the caller to keep track of how many files were downloaded vs how many were already cached. So I still think this is the cleanest solution, but I'm open to alternatives if you have one in mind.

banzai/tests/site_e2e/README.md
timbeccue: Readme appears to render fine here https://github.com/LCOGT/banzai/blob/ad67abdffbd0f1b474c3bf757edbf02514804e1d/banzai/tests/site_e2e/README.md

banzai/dbs.py
timbeccue: Hmm. What do you think of `StackSubframes` instead? 

The tricky part of using `FramesToStack` is that, since this also keeps track of successful and failed subframes, the db operation helper method names become a bit more awkward. Currently a query is `get_stack_frames`; if we change it to `get_frames_to_stack` it looks like we're pulling from a pool of pending work.

But I don't have a terribly strong opinion here, so tell me if you'd still like to go with FramesToStack and I can make that change.

banzai/tests/site_e2e/site_e2e.env.template
timbeccue: I have things working on both mac and linux, so I propose leaving this as-is (and not adding the complexity of creating a docker network) unless you still think it's worth doing now.

banzai/dbs.py
timbeccue: Table name has been updated to `Subframe` per further discussion in slack.

banzai/tests/site_e2e/test_site_e2e.py
timbeccue: I’ll make the default pytest invocation skip E2E-style tests via `addopts = -p no:warnings -m "not e2e and not e2e_site"`. New test commands:

```bash
# Unit tests, E2E skipped by default
uv run pytest

# Normal E2E tests
uv run pytest -m e2e banzai/tests/test_end_to_end.py -v

# Site E2E tests
uv run pytest -m e2e_site banzai/tests/site_e2e/test_site_e2e.py -v
```

banzai/tests/site_e2e/test_site_e2e.py
timbeccue: Fair point. The queue_images.py script was written to help me to manually enqueue stuff, but it's also used here as a test harness for putting a realistic message on the queue from outside the containers, to mimic site software. We could replace the subprocess call with a Python helper that publishes to RabbitMQ directly and keep the same deployment coverage. Would you like me to make that change?

banzai/tests/site_e2e/conftest.py
timbeccue: My intent is that site_e2e is a host-run deployment harness, not code that runs inside the banzai containers. Pytest is currently the outer orchestrator for Docker Compose. If we want the site_e2e tests to run in CI then I agree it makes sense to move the compose lifecycle management out to CI and let pytest talk only to the already running services. Are you ok if we defer that for now?

banzai/tests/site_e2e/conftest.py
timbeccue: Actually this specific function is unused and has been removed, but your question applies equally to `publish_raw_string_to_queue`. I've pulled that and `post_to_archive_queue` into a separate `utils/messaging.py`. 

I think `post_to_archive_queue` and `publish_raw_string_to_queue` are different enough (exchange type, use of routing key, is queue declared, body type, serialization) that keeping them separate is simpler for the codebase.

banzai/stacking.py
timbeccue: I'm not sure I understand what's motivating your question.

This function is run by the stack worker in a polling loop to check for notifications (in the redis queue) signifying new subframes that have been reduced and are ready for stacking. One of our goals is that as soon as the stack worker is free, it stacks all available frames, so in this function we're draining all the notifications that may have accumulated in the redis queue, rather than operating on them one by one. 

Maybe it would be cleaner to do something like this instead?
```python

def drain_notifications(redis_client, camera):
    key = f'{REDIS_KEY_PREFIX}{camera}'
    raw = []
    while True:
        item = redis_client.rpop(key)
        if item is None:
            break
        raw.append(item)
    return {item.decode() if isinstance(item, bytes) else item for item in raw}
```

Is there a better way we could do this?

banzai/scheduling.py
timbeccue: The stack worker is listening to redis to determine if we need to create a new stack. If redis contains entries indicating new subframes are ready, then we check the Subframe db table to get the relevant details and stack everything. So we're not really polling the db, but we are using it as the source of truth whenever a redis notification is available.

Is there a better way to be handling this?

banzai/stacking.py
timbeccue: The StackingSupervisor is in charge of managing the per-camera stacking subprocesses. But the actual worker function is outside the class, defined in run_worker_loop. Does this address your concern? I'll also make a process flow diagram.

banzai/cache/init.py
timbeccue: Changed from `site` to `site_deploy`

banzai/scheduling.py
timbeccue: What types of issues are you worried about when you say db polling is error prone? Is the current system of polling redis to check for new subframes exposing us to the same risk? 

I'm wondering whether it would be cleaner to remove the redis notification system entirely and just poll the db regularly, but then that's what you're warning against.

timbeccue:
Sacking architecture and process flow diagrams have been added here: https://github.com/LCOGT/banzai/pull/444/changes#diff-1df90f6630375c6d0b8ea4221090de028e494a40178e28d9a35f0458f6d82451

banzai/cache/download_worker.py
cmccully: Why does the caller care if the file is downloaded or already cached? Doesn't it just care that the file is on disk now?

banzai/tests/site_e2e/conftest.py
cmccully: I don't really think pytest should be in charge of container management. That feels like an anti-pattern to me. In principle we could use tox, but I'm not really sure what you are trying to accomplish, so I'm concerned.

banzai/tests/site_e2e/site_e2e.env.template
cmccully: I think this depends on how things are configured at site. If the other site containers are on a specific docker network then we should update to match.

banzai/tests/site_e2e/test_site_e2e.py
cmccully: I think that would be better here.

banzai/main.py
cmccully: This is still a useful comment.

banzai/stacking.py
cmccully: I think we should operate on files one by one in the same way that banzai already does. Then if something crashes you don't lose all the messages, just the most recent.

banzai/stacking.py
cmccully: It might be worth getting @jchate6 s opinion on the overall infrastructure. This is adding a lot of complexity to the deployment so we should make sure we have several eyes on it to make sure we understand how to fix things if something goes wrong.

banzai/cache/download_worker.py
timbeccue: Originally this was so that we could produce a log line at the end of a cache sync cycle that said something like: `X total files to cache; Y new files were downloaded`.

Resolved (via conversation): simply log how many files are needed to cache, and also log every new file downloaded. Then we can remove the return value from this function.

banzai/tests/site_e2e/conftest.py
timbeccue: I can refactor the test logic so that infrastructure setup and teardown happens via bash script, and pytest only runs tests on this already-running setup. Does that seem like it should be a separate PR though?

banzai/tests/site_e2e/site_e2e.env.template
timbeccue: Mark confirmed there's no pre-existing Docker network at site, I'll keep this as-is.

banzai/main.py
timbeccue: I'll add this refactor.

banzai/stacking.py
jchate6: I might not be understanding the full context, but I think it makes sense to run through all the messages, and just do a no-op if a message is for a frame that has already been stacked.

banzai/stacking.py
jchate6: I think it's definitely worth thinking about ways to simplify this infrastructure. However, it seems like this is basically ready to roll out. It's probably worth setting things up as is, testing and monitoring thoroughly and making sure normal operations aren't interrupted. If it works, and @timbeccue and @markBowman are willing to maintain it at site, I don't think it makes much sense to put the effort in to refactoring everything before the actual stacking code as even been written.

banzai/stacking.py
timbeccue: I've updated the Redis queue to be a sorted set. That simplifies things in two ways. First, we no longer have to manage multiple notifications for the same stack; the worker can pop stack A once and operate on all its available subframes. Second, the ordering lets us pop the oldest pending stack. If stack A is waiting for another subframe, the worker can process stack B; when A’s next subframe arrives, A re-enters the queue in arrival order. This prevents an incomplete stack from blocking other work without allowing newer notifications to jump ahead of older ones.
The worker processes one MOLUID completely before popping the next, so a crash leaves every unpopped notification intact.

cmccully (approved):

PR #445, opened 2026-03-20
Smartstack Scaffolding (replaced with #444)

timbeccue:
Add the infrastructure for smartstack observations. This includes the stacking core/data model and the
pipeline integration to receive and process subframes.

- Add StackFrame DB model and helper functions to track frames during smartstack observations
- Add stacking worker (Redis-based per-camera notifications, stack completion check, timeout detection)
- Add stacking supervisor (runs a separate stack worker per camera)
- Add SubframeListener (similar to banzai-listener)
- Add process_subframe worker (similar to banzai-worker, but also sends reduced subframes to the stack worker)
- Add banzai_subframe_worker entry point
- Modify run_pipeline_stages to return the reduced image filepath, for passing to the stack DB/worker
- Unit tests and e2e tests

These changes add everything needed for smartstacking except the stacking algorithm itself.

Note: stage_utils.py:run_pipeline_stages previously had no return value and now returns the processed
images. This shouldn't affect existing callers but is worth flagging since it modifies a core banzai
function.


timbeccue:
Closing this as it has been combined into #444

PR #449, opened 2026-03-26
Update README installation instructions for Poetry

timbeccue:
Replace pip-based setup with Poetry/pyenv workflow. 

Document Python 3.12 requirement (3.13 lacks numpy<2 wheels on Linux).


markBowman (approved):
Thank you

timbeccue:
Banzai has switched to uv from poetry, so these changes are no longer relevant. Closing.

PR #450, opened 2026-03-26
Suppress meaningless opentsdb metrics connection warning

timbeccue:
Set default `OPENTSDB_PYTHON_METRICS_TEST_MODE=1` in `banzai/__init__.py` so users don't see this confusing error on every run:

`ERROR: metric_wrappers: ConnectionError(MaxRetryError("HTTPConnectionPool(host='scheduler-dev.lco.gtn', port=80): Max retries exceeded with url: /api/version (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x112013230>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))"))`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 Versions
 ========
 
+1.33.2 (2026-04-07)
+-------------------
+
+- Suppress ingester warning for opentsdb if OPENTSDB_HOSTNAME env variable is not set.
+
 1.33.1 (2026-04-07)
 -------------------
 
--- a/banzai/__init__.py
+++ b/banzai/__init__.py
@@ -3,8 +3,12 @@
 # Packages may add whatever they like to this file, but
 # should keep this content at the top.
 # ----------------------------------------------------------------------------
+import os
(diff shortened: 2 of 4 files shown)


timbeccue:
The current solution only sets `OPENTSDB_PYTHON_METRICS_TEST_MODE` if we're not in prod because using any value at all evaluated to `True`. I made an update to [opentsdb-python-metrics](https://pypi.org/project/opentsdb-python-metrics/0.3.0/) so in version 0.3.0, setting that variable to '0' or 'false' correctly sets test mode to off. 

We could make this solution cleaner by having banzai's production deployment explicitly configure `OPENTSDB_PYTHON_METRICS_TEST_MODE` to be 'false', then also safely use a default of 'true' for all other cases. 

However, the ocs ingester needs to be updated to use opentsdb-python-metrics 0.3.0 first, and then banzai needs to specify that version of the ingester before we can access this change.

cmccully (approved):
Add a change log message and bump the last digit of the version. Otherwise looks good.

PR #454, opened 2026-04-07
Update psycopg to v3 to work with the defaults in sqlalchemy 2

cmccully:


pyproject.toml
```diff
@@ -33,7 +33,7 @@ dependencies = [
   "opensearch-py",
   "pytest>=4.0",
   "pyyaml",
-  "psycopg2-binary",
+  "psycopg[binary]",
```
sfoale, line 36: After upgrading psycopg to v3 I think you will need to change any postgres connection urls to postgresql+psycopg:// to target the psycopg3 driver.. See https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#module-sqlalchemy.dialects.postgresql.psycopg

sfoale (changes requested):
Not sure this is going to work with the existing postgres connection urls, if I'm reading the documentation right,

sfoale (approved):
Looks good now.

PR #455, opened 2026-04-08
Added alembic script to migrate instrument table from previous banzai versions

cmccully:


timbeccue (approved):
Alembic is new to me so I may not be the best reviewer, but I don't see any issues.

jchate6 (approved):
Seems scary, but I don't see anything obviously wrong.

PR #457, opened 2026-04-16
Fix astropy fpack

cmccully:
This PR will pull in fixes I made to astropy to fpack data correctly. We also have added more logging for exceptions in querying the archive.


jchate6 (approved):
Once this is deployed, we should follow the logs very closely and try to maybe follow a frame all the way through and make sure it's funpackable by the end?

PR #459, opened 2026-04-20
Add auth to obs-portal calls for calibration block schedule.

cmccully:
We do this for obs portal queries to schedule when we need to stack calibrations. This is now required by the obs portal in an attempt to block outside bots.


jchate6 (approved):
good catch in the logs.

PR #461, opened 2026-05-08
Added daily requeuing of missing frames.

cmccully:
This PR adds a cron to requeue missing frames using the same cron container as the calibration scheduler. The query only relies on the archive api.

I also refactored some of the archive querying for better reuse of the retry logic and less code duplication.

### DIFF
--- a/banzai/main.py
+++ b/banzai/main.py
@@ -54,12 +54,6 @@ class RealtimeModeListener(ConsumerMixin):
     def on_message(self, body, message):
         logger.info('Received message', extra_tags={'filename': body['filename']})
         try:
-            instrument = LCOFrameFactory.get_instrument_from_header(body, self.runtime_context.db_address)
-        except Exception:
-            logger.error(f'Could not get instrument from header. {traceback.format_exc()}', extra_tags={'filename': body['filename']})
-            message.ack()
-            return
-        try:
             queue_name = get_processing_queue(body, self.runtime_context)
         except Exception:
             message.ack()
@@ -67,7 +61,7 @@ class RealtimeModeListener(ConsumerMixin):
 
         process_image.apply_async(args=(body, vars(self.runtime_context)),
                                   queue=queue_name)
-        message.ack()  # acknowledge to the sender we got this message (it can be popped)
+        message.ack()
 
 
 def add_settings_to_context(args, settings):
@@ -377,7 +371,7 @@ def add_bpms_from_archive():
     # Query the archive for all bpm files
     url = f'{settings.ARCHIVE_FRAME_URL}/?OBSTYPE=BPM&limit=1000&include_related_frames=false'
     archive_auth_header = settings.ARCHIVE_AUTH_HEADER
-    response = archive_get(url, params={}, headers=archive_auth_header)
+    response = archive_get(url, params={}, auth_headers=archive_auth_header)
     results = response.json()['results']
 
     # Load each one, saving the calibration info for each
--- a/banzai/main.py
+++ b/banzai/main.py
@@ -188,12 +188,21 @@ def start_banzai_cron():
         app.add_periodic_task(crontab(minute=entry['minute'], hour=entry['hour']),
                               schedule_calibration_stacking.s(site=site, runtime_context=vars(runtime_context)),
                               queue=runtime_context.CELERY_TASK_QUEUE_NAME)
-    app.add_periodic_task(crontab(hour=runtime_context.REQUEUE_MISSING_FRAMES_TIME.hour,
-                                  minute=runtime_context.REQUEUE_MISSING_FRAMES_TIME.minute),
-                                  requeue_missing_frames.s(runtime_context=vars(runtime_context)),
-                                  queue=runtime_context.CELERY_TASK_QUEUE_NAME)
+
+        timezone = dbs.get_site(site, runtime_context.db_address).timezone
+
+        local_time_in_utc = date_utils.local_to_utc(
+            settings.REQUEUE_MISSING_FRAMES_TIME,
+            timezone
+        )
+        app.add_periodic_task(
+            crontab(hour=local_time_in_utc.hour, minute=local_time_in_utc.minute),
+            requeue_missing_frames.s(site, runtime_context=vars(runtime_context)),
+            queue=runtime_context.CELERY_TASK_QUEUE_NAME
+        )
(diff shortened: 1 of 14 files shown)


banzai/settings.py
```diff
@@ -163,3 +164,9 @@
 LARGE_WORKER_QUEUE = os.getenv('CELERY_LARGE_TASK_QUEUE_NAME', 'celery_large')
 
 REFERENCE_CATALOG_URL = os.getenv('REFERENCE_CATALOG_URL', 'http://phot-catalog.lco.gtn/')
+
+REQUEUE_MISSING_FRAMES_TIME = datetime.time(hour=14, minute=30)
+
+REQUEUE_OBSTYPES = ['EXPOSE', 'STANDARD']
+
+REQUEUE_LOOKBACK_HOURS = 36
```
jchate6: This feels like an odd lookback time. 
This means that everything from 02:30-14:30 is checked twice while everything from 14:30 to 02:30 is only checked once.

It isn't clear to me why we would have redundancy for OGG and ELP, but not CPT and TFN... 
It feels like this should be either 25 or 26 hours if we just want a little overlap to make sure we didn't miss anything during processing, or a full 48 hours if we want to check everything twice.

banzai/settings.py
```diff
@@ -163,3 +164,9 @@
 LARGE_WORKER_QUEUE = os.getenv('CELERY_LARGE_TASK_QUEUE_NAME', 'celery_large')
 
 REFERENCE_CATALOG_URL = os.getenv('REFERENCE_CATALOG_URL', 'http://phot-catalog.lco.gtn/')
+
+REQUEUE_MISSING_FRAMES_TIME = datetime.time(hour=14, minute=30)
```
jchate6: Are we worried about this being too close to the 15:00 report timing? If there are more than a few files, is it possible these might be missed?

banzai/main.py
```diff
@@ -184,14 +181,17 @@ def make_master_calibrations():
...
     logger.info('Entered entrypoint to celery beat scheduling')
     runtime_context = parse_args(settings)
     for site, entry in runtime_context.SCHEDULE_STACKING_CRON_ENTRIES.items():
         app.add_periodic_task(crontab(minute=entry['minute'], hour=entry['hour']),
                               schedule_calibration_stacking.s(site=site, runtime_context=vars(runtime_context)),
                               queue=runtime_context.CELERY_TASK_QUEUE_NAME)
-
+    app.add_periodic_task(crontab(hour=runtime_context.REQUEUE_MISSING_FRAMES_TIME.hour,
+                                  minute=runtime_context.REQUEUE_MISSING_FRAMES_TIME.minute),
+                                  requeue_missing_frames.s(runtime_context=vars(runtime_context)),
```
jchate6: I'm not sure what the `.s` is doing... is that a celery thing?

jchate6 (changes requested):
I would like to see more documentation here if possible.

- New or heavily reworked functions would benefit from doc strings describing their purpose, use, and intended functionality.

- I think we need to record this change of behavior somewhere easily accessible and referenced so that the precise details of Banzai's expected behavior are understood by a wider audience.

- There are several changes here that I don't really understand why they were made (like the changes in `fits_utils`). Some comments in the PR explaining how these different changes are related would generally be appreciated. 

Also, there are no test changes. Do we want to test that we catch and re-queue  missing frames correctly?

I'm curious if we have a different strategy for the cron. Does it make more sense to run it more than once a day?  What if we triggered for each site at local noon rather than one bulk re-queue in the West Coast morning?  This would spread out the queries and queuing as well.

banzai/scheduling.py
```diff
@@ -247,3 +252,40 @@ def process_image(self, file_info: dict, runtime_context: dict):
...
+def requeue_missing_frames(runtime_context: dict):
+    try:
+        runtime_context = Context(runtime_context)
+        logger.info('Checking for missing frames to requeue')
+        raw_frames = []
+        reduced_frames = []
+        for obstype in runtime_context.REQUEUE_OBSTYPES:
+            # Get the raw frames that we took
+            start = datetime.now(timezone.utc) - timedelta(hours=runtime_context.REQUEUE_LOOKBACK_HOURS)
+            end = datetime.now(timezone.utc)
```
jchate6, line 278: Is there any concern here about re-queuing things that are just in the middle of being processed?

banzai/query.py
```diff
@@ -0,0 +1,67 @@
...
+        if response.json()['next'] is None:
+            more_frames = False
+        else:
+            logger.debug(f"Getting more {obstype} frames. So far we have {len(frames)} frames.")
+            response = archive_get(response.json()['next'], {}, auth_headers=auth_headers)
+            frames += response.json()['results']
+    return frames
+
+
+def cross_match_missing_frames(raw_frames, reduced_frames):
```
jchate6, line 107: This will also consider any unbinned UBVRI frames as missing and re-queue them, correct?

cmccully: If they are missing then yes, but it will not automatically requeue frames that already processed even if they are unbinned.

jchate6: We don't process those at all unless specifically requested.
There are many warnings and it hasn't happened in months, but they will always be unprocessed.

banzai/settings.py
cmccully: I'll change this to 48 hours so we check everything twice. I'll also modify things to check by site at local noon.

banzai/settings.py
cmccully: This should be addressed by changing to site noon. If we notice one site continues to have issues, we can use like 10 am local or something.

banzai/main.py
cmccully: This is the standard celery function to make a signature object which is how cron jobs are registered. 
https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html#entries

banzai/scheduling.py
cmccully: Not really. We'll end up with an extra file in the archive but the odds of this should be tiny. Especially if we run at local noon.

banzai/query.py
cmccully: So this will resubmit them, but then they will just decide to not process them as they did the first time around.

cmccully:
The archive_get function doesn't change any of the behavior. It is just a refactor of previously duplicated retries. I've added a docstring.

jchate6 (approved):
Looks good. Well done!

PR #463, opened 2026-05-16
Add timeout check for smartstacks

timbeccue:
## Summary

Adds fixed timeout handling for smart-stack subframes.

The basic behavior is that an in-progress smart stack enters a timeout state if more than 15 minutes elapses between reduced subframes. This will trigger the final stack and cleanup routine for the smartstack (not yet implemented). The purpose is to prevent a hanging stack from indefinitely blocking the stack worker from picking up subsequent stack jobs. 

## Details

- Adds fixed latest-reduced-row timeout detection for active smart stacks
- Marks incomplete stale stacks as `timeout`
- Keeps complete stacks marked as `complete` before considering timeout
- Makes `get_subframes()` return rows ordered by `stack_num`
- Keeps timeout bookkeeping based only on reduced subframe rows with required filepaths
- Updates smart-stacking tests and architecture docs for the fixed timeout behavior

### DIFF
--- a/banzai/dbs.py
+++ b/banzai/dbs.py
@@ -139,7 +139,6 @@ class Subframe(SiteBase):
     is_last = Column(Boolean, default=False)
     status = Column(String(20), default='active', nullable=False)
     dateobs = Column(DateTime, nullable=True)
-    exptime = Column(Float, nullable=True)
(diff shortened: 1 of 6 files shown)


timbeccue:
Will be covered in a different PR.

PR #464, opened 2026-05-26
Fix to local to utc that only worked with datetimes and not time objects.

cmccully:
We were passing a datetime.time object into the time conversion for requeuing frames, but the function we were calling only works with datetime.dateime objects. This fixes that.


jchate6 (approved):
pending tests...

PR #465, opened 2026-05-27
Remove related frames from query that doesn't use it

jnation3406:
This is what I messaged you about last week - I don't think the related frames are needed from this query so its best to not include them.

Can someone more familiar with banzai tell me why the tests are failing?


cmccully (approved):
Looks good.

PR #468, opened 2026-06-23
Fix to tests for requeuing frames.

cmccully:
You can also note in query.py to line 133 that got prematurely merged.


jchate6 (approved):
yay tests!

PR #469, opened 2026-07-16
Smartstack - image stacking

timbeccue:
This is the first of four PRs that can be reviewed independently in service of smartstacking now that the scaffolding is in place.

The calibration stacking code has been refactored so that it can be reused for smartstacking as well. Added method="sum" for smartstacks, in addition to the existing mean capabilities used for calibrations. 

I've included regression coverage with the prior implementation in pre_refactor_stack_snapshot to demonstrate that behavior for calibration frames has not changed. Is this useful to keep around long term? 

I have _warn_if_exposure_times_differ for all stacks, which issues a warning if exposures times differ by more than 1%. If calibration frames use different exposure times then I can have this only run for smartstacks.

### DIFF
--- a/banzai/data.py
+++ b/banzai/data.py
@@ -240,17 +240,21 @@ class CCDData(Data):
         inner_ny = round(self.data.shape[0] * inner_edge_width)
         return self.data[inner_ny: -inner_ny, inner_nx: -inner_nx]
 
-    def trim(self, trim_section=None):
+    def trim(self, trim_section=None, memmap=None):
         """
         :param trim_section: Always in data coords
+        :param memmap: Override whether to copy the trimmed arrays into new memory maps.
+                       Defaults to the source object's memory-map preference.
         :return:
         """
         if trim_section is None:
             trim_section = Section.parse_region_keyword(self.meta.get('TRIMSEC', 'N/A'))
+        if memmap is None:
+            memmap = self.memmap
 
         trimmed_image = type(self)(data=self.data[trim_section.to_slice()], meta=self.meta,
                                    mask=self.mask[trim_section.to_slice()], name=self.name,
-                                   uncertainty=self.uncertainty[trim_section.to_slice()], memmap=self.memmap)
+                                   uncertainty=self.uncertainty[trim_section.to_slice()], memmap=memmap)
         trimmed_image.detector_section = self.data_to_detector_section(trim_section)
         trimmed_image.data_section = Section(x_start=1, y_start=1,
                                              x_stop=trimmed_image.data.shape[1],
@@ -521,36 +525,16 @@ def stack(data_to_stack, nsigma_reject, method='mean') -> CCDData:
         stacked_data = a.sum(axis=0) * n_images / n_good_pixels
         stacked_uncertainty = np.sqrt(variances.sum(axis=0)) * n_images / n_good_pixels
     else:
-        raise ValueError(f'Unknown stack method: {method}')
+        raise ValueError(f"Unknown stack method: {method!r}; expected 'mean' or 'sum'")
 
     return CCDData(data=stacked_data, meta=data_to_stack[0].meta, uncertainty=stacked_uncertainty, mask=stacked_mask)
-
-
-def science_hdu(frame):
-    """Return the science CCDData HDU for a frame or CCDData input.
-
-    Parameters
-    ----------
-    frame : CCDData or ObservationFrame
-        Input CCD data or frame containing CCD data.
-
-    Returns
-    -------
-    CCDData
-        Primary HDU when it contains CCD data, otherwise the first CCD extension.
-    """
-    if isinstance(frame, CCDData):
-        return frame
-    if isinstance(frame.primary_hdu, CCDData):
-        return frame.primary_hdu
-    return frame.ccd_hdus[0]
 
 
 def _warn_if_exposure_times_differ(images):
     """Log a warning when input exposure times differ by more than one percent."""
     exposure_times = []
     for image in images:
-        exptime = science_hdu(image).meta.get('EXPTIME')
+        exptime = image.meta.get('EXPTIME')
         if exptime is None:
             continue
         try:
@@ -572,15 +556,15 @@ def _warn_if_exposure_times_differ(images):
                        extra_tags={'exptime_spread_fraction': exptime_spread_fraction})
 
 
-def combine_images(images, output_frame, nsigma=3.0, method='mean'):
-    """Combine images into an output frame in memory-bounded y sections.
+def combine_images(images, output_image, nsigma=3.0, method='mean'):
+    """Combine CCD data into an output image in memory-bounded y sections.
 
     Parameters
     ----------
-    images : list of CCDData or ObservationFrame
-        Input images to combine.
-    output_frame : CCDData or ObservationFrame
-        Frame or CCD data object whose science HDU receives the combined result.
+    images : list of CCDData
+        Input CCD data to combine.
+    output_image : CCDData
+        CCD data object that receives the combined result.
     nsigma : float, optional
         Robust-standard-deviation threshold for rejecting pixels from each section.
     method : {'mean', 'sum'}, optional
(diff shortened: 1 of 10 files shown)


banzai/data.py
```diff
@@ -458,11 +479,148 @@ def stack(data_to_stack, nsigma_reject) -> CCDData:
...
+        n_images = len(data_to_stack)
+        stacked_data = a.sum(axis=0) * n_images / n_good_pixels
+        stacked_uncertainty = np.sqrt(variances.sum(axis=0)) * n_images / n_good_pixels
+    else:
+        raise ValueError(f'Unknown stack method: {method}')
 
     return CCDData(data=stacked_data, meta=data_to_stack[0].meta, uncertainty=stacked_uncertainty, mask=stacked_mask)
+
+
+def science_hdu(frame):
```
cmccully: This feels dangerous to always assume zero. Why not use the 'SCI' extension like everywhere else in the code? Also, why not just pass this directly? Why do we need a new function?

banzai/data.py
```diff
@@ -420,8 +423,26 @@ def background(self, value):
         self.data -= self._background
 
 
-def stack(data_to_stack, nsigma_reject) -> CCDData:
-    """
+def _apply_sigma_rejection(data_to_stack, nsigma_reject):
```
cmccully, line 430: Is this all array api (numpy 2) compatible? On first glance, it looks like we lose some of the parallelization in this PR.

banzai/data.py
```diff
@@ -458,11 +479,148 @@ def stack(data_to_stack, nsigma_reject) -> CCDData:
...
+
+    Returns
+    -------
+    CCDData or ObservationFrame
+        The same output object passed in, with combined science data copied into it.
+    """
+    _warn_if_exposure_times_differ(images)
+
+    # Turn off memory mapping for each segment.
+    for image in images:
```
cmccully: I don't think this is actually what you want to do. I think you want things memmaped, because otherwise this will read all the images back into memory.

banzai/data.py
```diff
@@ -458,11 +479,148 @@ def stack(data_to_stack, nsigma_reject) -> CCDData:
...
+                break
+            y_stop = input_hdu.shape[0]
+        else:
+            y_stop = (i + 1) * y_step
+
+        section_to_stack = Section(x_start=1, x_stop=input_hdu.data.shape[1], y_start=y_start, y_stop=y_stop)
+        data_to_stack = [science_hdu(image)[section_to_stack] for image in images]
+        stacked_data = stack(data_to_stack, nsigma, method=method)
+
+        output_hdu.copy_in(stacked_data)
```
cmccully: This feels like it needs a section keyword? Why not just do a output_hdu.data[:, :] = stacked_data[:,:]

banzai/tests/utils.py
```diff
@@ -54,6 +55,53 @@ def __init__(self, image_multiplier=1.0, nx=101, ny=103, name='test_image', read
             setattr(self, keyword, kwargs[keyword])
 
 
+def pre_refactor_stack_snapshot(data_to_stack, nsigma_reject):
```
cmccully: Regression tests make sense here when you are doing a refactor, but I don't know that they need to be long-lived. It's better to test the output matches rather than codifying past behavior. That's kind of an antipattern.

banzai/tests/test_stacking.py
```diff
@@ -1,12 +1,26 @@
...
-from banzai.data import stack
-from banzai.tests.utils import FakeCCDData
+from banzai import data as data_module
+from banzai.data import CCDData, stack
+from banzai.tests.utils import FakeCCDData, pre_refactor_stack_snapshot
 
 pytestmark = pytest.mark.stacking
 
 
+def _make_ccd(data, mask=None, uncertainty=None):
```
cmccully: I might rename this to make_cdd_data or something. You aren't really making a ccd.

banzai/tests/test_calibration_stacker.py
```diff
@@ -69,3 +71,28 @@ def test_stacking_with_different_pixels():
     np.testing.assert_allclose(stacked_data.data, d * np.mean(np.arange(9)))
     np.testing.assert_allclose(stacked_data.primary_hdu.uncertainty, np.ones((ny, nx)))
     assert np.all(stacked_data.mask == 0)
+
+
+def test_calibration_stacker_matches_pre_refactor_snapshot():
```
cmccully: Same discussion as below. I don't think you want to hard code in a past implementation.

cmccully:
Overall, it's a good start, but I worry we've given away parallelization and memory mapping for large numbers of images. I also wouldn't codify the old implementation in tests that run on every pr going forward. They are good smell tests for now, but don't need to be in the repo.

banzai/data.py
timbeccue: I think the full source arrays are already memmapped when CCDData is constructed. Setting image.memmap = False here didn’t convert them back into in-memory arrays, it prevented each sliced section from being copied into an additional temporary memmap, so the sections remained views of the existing source memmaps.
However, mutating the flag on the input images was probably confusing and could affect later use. I changed trim() to accept a per-call memmap override, and combine_images() now uses trim(..., memmap=False). Now, the full inputs remain memmapped and each section remains a view of its source.

banzai/data.py
timbeccue: stacked_data is the CCDData object that contains the slice of data we're currently stacking and the corresponding mask and uncertainty slices. With copy_in, the section of the data we stacked is placed in the correct output location based on the section's DETSEC, but your comment makes me realize that it would be better to do this more explicitly (which I've added).

banzai/data.py
```diff
@@ -458,11 +479,148 @@ def stack(data_to_stack, nsigma_reject) -> CCDData:
...
+    method : {'mean', 'sum'}, optional
+        Combine mode passed through to :func:`stack`.
+
+    Returns
+    -------
+    CCDData or ObservationFrame
+        The same output object passed in, with combined science data copied into it.
+    """
+    _warn_if_exposure_times_differ(images)
+
```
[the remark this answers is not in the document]
timbeccue: Changed to only call this warning on combinations with method="sum".

banzai/data.py
timbeccue: Updated to use SCI directly

banzai/data.py
```diff
@@ -458,11 +479,148 @@ def stack(data_to_stack, nsigma_reject) -> CCDData:
...
+    if method == 'mean':
+        stacked_data = a.sum(axis=0) / n_good_pixels
+        stacked_uncertainty = np.sqrt(variances.sum(axis=0) / (n_good_pixels ** 2.0))
+    elif method == 'sum':
+        n_images = len(data_to_stack)
+        stacked_data = a.sum(axis=0) * n_images / n_good_pixels
+        stacked_uncertainty = np.sqrt(variances.sum(axis=0)) * n_images / n_good_pixels
+    else:
+        raise ValueError(f'Unknown stack method: {method}')
```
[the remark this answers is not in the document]
timbeccue: Added this

banzai/tests/utils.py
timbeccue: Good point; removed

banzai/tests/test_stacking.py
timbeccue: 👍

banzai/tests/test_stacking.py
```diff
@@ -1,12 +1,26 @@
 import numpy as np
 import pytest
+from astropy.io import fits
 
-from banzai.data import stack
-from banzai.tests.utils import FakeCCDData
+from banzai import data as data_module
+from banzai.data import CCDData, stack
+from banzai.tests.utils import FakeCCDData, pre_refactor_stack_snapshot
```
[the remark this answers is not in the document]
timbeccue: removed

banzai/data.py
timbeccue: Not sure I completely understand your comment. If you're asking about numpy 2 compatibility, it passes the stacking and median tests on numpy 2.4.4.

In terms of parallelism, _apply_sigma_rejection() is the existing body of stack() extracted into a helper. Both median calculations still go through the same cython implementation, and the section loop remains sequential as it was before. The only new numerical path is the final sum mode.

Can you be more specific about your concern?

cmccully (approved):

PR #471, opened 2026-07-21
Smartstack: db schema

timbeccue:
## Summary

This is PR 3 of 8 completing the v1 banzai implementation of Smartstacking at site.

PR1 (#469): combine math
PR2 (#470): JPEG utilities
PR3 (#471): DB updates   &emsp;&emsp;&emsp;&emsp;    <-- you are here
PR4 (#473): shipper contract                     
PR8 (#477): logging

```text
PR1 combine math ───────┐
PR2 JPEG utilities ─────┤
PR4 shipper contract ───┘
```

PRs 1–4 provide the independent foundations for combining images, rendering JPEGs, storing stack state (this PR), and publishing products. PR5 builds the products, PR6 adds the polling worker, and PR7–8 add end-to-end proof, logging, and final documentation.

This PR owns one handoff:

> successful stackframe reduction → durable PostgreSQL state

## What changes here

The database schema has been changed from a single `subframes` table to now having a `stack` and `stackframes` table. 

The renaming of 'subframes' to 'stackframes' was done to make the relationship between a stack and a stackframe more obvious. 

## Why change the db schema?

1. I wanted to get rid of the awkward way we stored stack state in the subframes table. Each row represented a single frame, but repeated info like the stack status or completion time across each row. Having a one-to-many relationship between the stack and stackframes better fits the data.
2. My takeaway from the conversation with Joey and Daniel was that the stack worker polling was overly complex. We had stack workers polling redis to detect whether to read the database to check for new frames. The new stack row makes each Smartstack an explicit, discoverable unit of work. This lets the later worker poll durable state directly and remain stateless across restarts, without maintaining a second Smartstack notification path.

## Scope

This PR stops at the database handoff. Product generation belongs to PR5, while polling and finalization belong to PR6.

### `stacks`

| Column | Type | Constraints / default |
|---|---|---|
| `moluid` | `String(100)` | Primary key |
| `camera` | `String(50)` | Indexed, nullable |
| `frmtotal` | `Integer` | Nullable |
| `status` | `String(20)` | Nullable; default: `active` |
| `created_at` | `DateTime` | Nullable; default: current UTC time |
| `last_stackframe_at` | `DateTime` | Nullable; default: current UTC time |
| `completed_at` | `DateTime` | Nullable |
| `finalize_attempts` | `Integer` | Nullable; default: `0` |
| `next_attempt_at` | `DateTime` | Nullable |
| `last_preview_count` | `Integer` | Nullable; default: `0` |

### `stackframes`

| Column | Type | Constraints / default |
|---|---|---|
| `id` | `Integer` | Primary key, auto-incrementing |
| `moluid` | `String(100)` | Foreign key to `stacks.moluid`, indexed, nullable; cascades on delete |
| `stack_num` | `Integer` | Nullable |
| `filepath` | `String(255)` | Not nullable |
| `dateobs` | `DateTime` | Nullable |
| `is_last` | `Boolean` | Nullable; default: `false` |
| `created_at` | `DateTime` | Nullable; default: current UTC time |

The (`moluid`, `stack_num`) pair is unique.


banzai/alembic/versions/9a4e2f7c1d3b_rework_smartstack_schema.py
```diff
@@ -0,0 +1,92 @@
...
+    # These tables are also created directly from the SQLAlchemy models by create_db(site_deploy=True),
+    # so this migration must tolerate either starting state.
+    if inspector.has_table('subframes'):
+        op.drop_table('subframes')
+    if not inspector.has_table('stacks'):
+        op.create_table(
+            'stacks',
+            sa.Column('moluid', sa.String(length=100), nullable=False),
+            sa.Column('camera', sa.String(length=50), nullable=True),
+            sa.Column('frmtotal', sa.Integer(), nullable=True),
```
cmccully, line 34: This is mostly taste, but what do you get by removing 2 letters here? moluid is from the headers so not such a big deal, but this took mental effort to figure out which frm (form, from, frame,...) we were talking about.

banzai/dbs.py
```diff
@@ -128,21 +130,32 @@ class ProcessedImage(Base):
...
+    status = Column(String(20), default='active')
+    created_at = Column(DateTime, default=datetime.datetime.utcnow)
+    last_stackframe_at = Column(DateTime, default=datetime.datetime.utcnow)
+    completed_at = Column(DateTime, nullable=True)
+    finalize_attempts = Column(Integer, default=0)
+    next_attempt_at = Column(DateTime, nullable=True)
+    last_preview_count = Column(Integer, default=0)
+
+
+class Stackframe(SiteBase):
```
cmccully, line 147: This kind of feels like it should be camel case. Check pep8 for naming conventions.

banzai/dbs.py
```diff
@@ -624,63 +637,146 @@ def replicate_instrument(instrument_record, db_address):
         db_session.commit()
 
 
-def insert_subframe(db_address, moluid, stack_num, frmtotal, camera, filepath, is_last, dateobs):
-    """Upsert a subframe record. On conflict (moluid, stack_num), reset the row
-    to a fresh active state so a requeue behaves like a retry from scratch."""
+def _insert_for_session(session):
```
cmccully, line 640: Why are we not using normal sqlalchemy inserts here? This feels like we should just use the ORM.

cmccully (approved):
Some minor nits, but looks fine.

banzai/dbs.py
timbeccue: We want an insert-like behavior that still works if an entry already exists. The ORM’s merge() looks up rows by id, but here the frame is identified by (moluid, stack_num). So we use ON CONFLICT so duplicate deliveries update the same stackframe instead of failing.

banzai/alembic/versions/9a4e2f7c1d3b_rework_smartstack_schema.py
timbeccue: I'm referencing frmtotal from the headers as well, but I'll add a comment to clarify the meaning.

banzai/dbs.py
timbeccue: I agree camel case is appropriate here, but I'm using stackframe as a single word word (like subframe).

banzai/scheduling.py
```diff
@@ -328,16 +326,16 @@ def process_subframe(self, body: dict, runtime_context: dict):
...
             images[0].get_output_directory(runtime_context),
             images[0].get_output_filename(runtime_context),
         )
-        # Phase 2: Record reduced subframes in the Subframes table: ready to be stacked
-        insert_subframe(
+        # Phase 2: Record reduced stackframes in the DB; the stacking worker polls from there.
+        stack_created = upsert_stack_and_stackframe(
             runtime_context.db_address,
             moluid=header['MOLUID'],
             stack_num=header['MOLFRNUM'],
```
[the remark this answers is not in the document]
timbeccue: This PR is intended to work in conjunction with #475, which modifies the stacking supervisor behavior to poll the db directly rather than use redis notifications.

PR #473, opened 2026-08-06
Smartstack: shipper queue

timbeccue:
Summary

This is PR 4 of 8 completing the v1 banzai implementation of Smartstacking at site.

PR1 (#469): combine math
PR2 (#470): JPEG utilities
PR3 (#471): DB updates   
PR4 (#473): shipper contract         &emsp;&emsp;&emsp;&emsp;    <-- you are here            
PR5 (#474): smartstack data products
PR6 (#475): stack worker polling
PR7 (#476): e2e tests, compose cleanup
PR8 (#477): logging

```text
PR1 combine math ───────┐
PR2 JPEG utilities ─────┤
PR3 DB updates ─────────┼──> PR5 data products --> PR6 stack workers --> PR7 E2E --> PR8 logs/docs
PR4 shipper contract ───┘
```

PRs 1–4 provide the independent foundations for combining images, rendering JPEGs, storing stack state, and shipper integration (this PR). PR5 builds the products, PR6 adds the polling worker, and PR7–8 add end-to-end proof, logging, and final documentation.

## What changes here

Added methods to post data to the shipper queue. In the case of smartstack previews, this includes passing in a subset of the fits header so that the ingester has the metadata it needs without the inclusion of a fits file. 

## Publishing contract

`post_to_shipper_queue()` publishes UTF-8, plain-text JSON:

| Field | Final product | Preview |
|---|---|---|
| `fits` | `e45` FITS path | `null` |
| `small_thumbnail` | JPEG path | JPEG path |
| `large_thumbnail` | JPEG path | JPEG path |
| `instrument_enqueue_timestamp` | Created at queue insert time | Created at queue insert time |
| `thumbnail_metadata` | Omitted by the current final caller | Included when supplied |

The publisher also:

- Declares the durable fanout exchange, queue, and binding before publishing.
- Uses publisher confirms so broker rejection or acknowledgement timeout reaches PR6’s retry path.
- Applies bounded five-second connection and confirmation timeouts.
- Publishes and declares through the same channel.

New settings provide these defaults:

- `SHIPPER_EXCHANGE=ship_files`
- `SHIPPER_QUEUE_NAME=ship`
- `SMARTSTACK_PREVIEWS=true`

## Scope

This PR defines and tests the publishing boundary only. It does not create products or decide when to preview, finalize, or retry; those responsibilities belong to PR5 and PR6.

Focused tests cover the durable binding, plain-text JSON bodies for previews and finals, optional thumbnail metadata, publisher confirms, and timeout options.

PR #474, opened 2026-08-06
Smartstack: build data products (e45 fits and jpgs)

timbeccue:
## Summary

This is PR 5 of 8 completing the v1 banzai implementation of Smartstacking at site.

PR1 (#469): combine math
PR2 (#470): JPEG utilities
PR3 (#471): DB updates  
PR4 (#473): shipper contract                     
PR5 (#474): smartstack data products  &emsp;&emsp;&emsp;&emsp;    <-- you are here
PR8 (#477): logging

```text
PR1 combine math ───────┐
PR2 JPEG utilities ─────┤
PR4 shipper contract ───┘
```

PRs 1–4 provide the independent foundations for combining images, rendering JPEGs, storing stack state, and shipper integration (this PR). This PR brings them together into the product layer:

> ordered reduced `e09` stackframes → preview JPEGs or a final `e45` FITS product

PR6 will own the lifecycle decisions (stack worker behavior) that call this layer.

Note on merging: this should be merged into main only after PR1-4 are merged. Right now the base is set to simulate that so the diffs are easy to read, but the base should be changed to main before it's actually merged.

## What the changes do:

- Opens every persisted stackframe; fails rather than silently building from fewer inputs.
- Sorts members by `stack_num`.
- Creates a fresh output frame without modifying an input.
- Calls the image stacking code from PR1
- Derives the stack filename from the lowest-numbered available member by replacing `e09` with `e45`.
- Renders 300-pixel and 900-pixel JPEG derivatives using PR2’s utilities.

The final FITS records aggregate and provenance metadata in both the primary and `SCI` headers, including total `EXPTIME`, earliest `DATE-OBS`, `NCOMBINE`, `MOLUID`, and ordered `IMCOMnnn` inputs. It also removes per-exposure `MOLFRNUM`, scales stack-dependent detector values, and remeasures the combined-image background statistics.

## Header changes in stacked products:

While most fits headers are copied over from the stackframes, some are recalculated. I think the procedures make sense, but they're definitely worth a close review. 

- EXPTIME: sum of all input exposure times (matches the renormalized-sum flux scale).
- DATE-OBS: earliest input's observation start time.
- UTSTOP: copied from the newest input, so DATE-OBS/UTSTOP bracket the stack's full time span.
- SATURATE: inherited value × N (number of inputs); exact upper bound for trustworthy pixel values on the N-frame flux scale.
- MAXLIN: inherited value × N; same argument for the linearity limit.
- RDNOISE: inherited value × √N; read-noise variances add across a sum.
- L1MEAN / L1MEDIAN / L1SIGMA: remeasured on the stacked SCI using the same Background2D background-map method as SourceDetector, so they're comparable to e09 values.

## Preview and final paths

`run_preview()`:

- Builds the current stack in memory.
- Writes JPEGs but no FITS product.
- Publishes through PR4 with `fits: null` and aggregate thumbnail metadata.

`run_final()`:

- Writes the `e45` FITS and both JPEGs.
- Returns the actual written paths.
- Does not publish or change database state; PR6 owns that ordering.

## Scope

The PR-local implementation is `banzai/smartstack_products.py` and its focused tests. The combine, JPEG, database, and publisher layers visible in the cumulative branch are inherited from PR1–4.

This PR builds products but does not poll for work or manage lifecycle state. Tests cover output structure, metadata and provenance, stable naming, invalid inputs, preview publication, and a real FITS write/reopen path.

### DIFF
--- a/banzai/smartstack_products.py
+++ b/banzai/smartstack_products.py
@@ -119,13 +119,20 @@ def apply_smartstack_metadata(output_frame, input_images, stackframes, moluid):
     if science_header is not primary_header:
         headers.append(science_header)
 
+    # The stack is a sum renormalized per-pixel by N/n_good (see data.stack), so every pixel is on an
+    # N-frame-equivalent flux scale: N * SATURATE/MAXLIN are exact upper bounds for trustworthy values,
+    # and summed read-noise variances give sqrt(N) * RDNOISE.
     n_stacked = len(sorted_inputs)
     keyword_scale_factors = {'SATURATE': n_stacked, 'MAXLIN': n_stacked, 'RDNOISE': np.sqrt(n_stacked)}
+    # UTSTOP from the newest input so DATE-OBS (earliest) and UTSTOP bracket the full time span.
     newest_image = sorted_inputs[-1][1]
     utstop = newest_image['SCI'].meta.get('UTSTOP')
     if utstop is None:
         utstop = newest_image.primary_hdu.meta.get('UTSTOP')
 
+    # The inherited L1MEAN/L1MEDIAN/L1SIGMA describe the first input, not the stack. Remeasure with the
+    # same background-map method as photometry.SourceDetector so e45 values are comparable to e09 ones
+    # (raw-pixel stats would inflate L1SIGMA with pixel noise; the map measures background variation).
     background = Background2D(output_frame['SCI'].data, BACKGROUND_BOX_SIZE,
                               filter_size=BACKGROUND_FILTER_SIZE).background
     mean_background = stats.sigma_clipped_mean(background, BACKGROUND_NSIGMA_CLIP)
@@ -138,6 +145,7 @@ def apply_smartstack_metadata(output_frame, input_images, stackframes, moluid):
         header['NCOMBINE'] = (n_stacked, 'Number of images combined')
         header['MOLUID'] = (moluid, 'Observation request UID')
         header['DATE'] = (date_utils.date_obs_to_string(date_created), '[UTC] Date this FITS file was written')
+        # MOLFRNUM is a per-exposure sequence number; the combined product has no single position.
         if 'MOLFRNUM' in header:
             del header['MOLFRNUM']
         for keyword, scale_factor in keyword_scale_factors.items():
--- a/banzai/smartstack_products.py
+++ b/banzai/smartstack_products.py
@@ -3,7 +3,7 @@ import os
 
 import numpy as np
 
-from banzai.data import CCDData, HeaderOnly, combine_images, science_hdu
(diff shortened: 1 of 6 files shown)


cmccully:
Does this follow the filename convention we decided on? Did we end up agreeing that we would just choose the first filename number to be the final filename with e45?

banzai/smartstack_products.py
```diff
@@ -0,0 +1,243 @@
...
+
+def _zeroed_ccd_hdu(ccd_hdu):
+    return type(ccd_hdu)(
+        data=np.zeros_like(ccd_hdu.data),
+        meta=ccd_hdu.meta.copy(),
+        name=ccd_hdu.name,
+        mask=np.zeros_like(ccd_hdu.mask),
+        uncertainty=np.zeros_like(ccd_hdu.uncertainty),
+        memmap=ccd_hdu.memmap,
+    )
```
cmccully: Pretty much everything above this seems like duplicated code from other places in banzai. Should we just reuse code that's already written?

banzai/smartstack_products.py
```diff
@@ -0,0 +1,243 @@
...
+    # UTSTOP from the newest input so DATE-OBS (earliest) and UTSTOP bracket the full time span.
+    newest_image = sorted_inputs[-1][1]
+    utstop = newest_image['SCI'].meta.get('UTSTOP')
+    if utstop is None:
+        utstop = newest_image.primary_hdu.meta.get('UTSTOP')
+
+    # The inherited L1MEAN/L1MEDIAN/L1SIGMA describe the first input, not the stack. Remeasure with the
+    # same background-map method as photometry.SourceDetector so e45 values are comparable to e09 ones
+    # (raw-pixel stats would inflate L1SIGMA with pixel noise; the map measures background variation).
+    background = Background2D(output_frame['SCI'].data, BACKGROUND_BOX_SIZE,
```
cmccully: It feels like this should be refactored so here and in the background stage call the same code.

timbeccue: done

banzai/smartstack_products.py
timbeccue: I've extracted a shared helper, `create_combination_output_hdu`, for creating the blank SCI HDU for combine_images() to fill. 

For image opening, the similar code looks to be in stage_utils.run_pipeline_stages(). However, in that function failed images need to be skipped, while in the case of smartstacks, we want failed images to raise immediately with the failed filepath (to not create an incomplete stack). So a common helper would either need a bunch of options like skip_missing, fail_on_missing, and custom error/path handling, or only construct the frame factory and call open() which only saves us a couple lines. Maybe not worth it in this case?

cmccully (approved):

PR #475, opened 2026-08-06
Smartstack: stack worker and supervisor

timbeccue:
## Summary

This is PR 6 of 8 completing the v1 banzai implementation of Smartstacking at site.

PR1 (#469): combine math
PR2 (#470): JPEG utilities
PR3 (#471): DB updates   
PR4 (#473): shipper contract                     
PR6 (#475): stack worker polling   &emsp;&emsp;&emsp;&emsp;    <-- you are here
PR8 (#477): logging

```text
PR1 combine math ───────┐
PR2 JPEG utilities ─────┤
PR4 shipper contract ───┘
```

PRs 1–4 provide the independent foundations for combining images, rendering JPEGs, storing stack state, and shipper integration (this PR). PR5 builds the data products, this PR adds the polling worker, crash-only supervisor, and site deployment configuration. PR7 adds cross-service E2E and recovery proof; PR8 adds structured lifecycle telemetry and final documentation.

## Stack Worker Lifecycle behavior

The runtime starts one stateless polling process per camera. Each poll:

- Loads active stacks and their ordered members from PostgreSQL.
- Finalizes a stack when all expected members have arrived or an `is_last` member is present.
- Gives completion priority when a stack is both complete and past its timeout.
- Defines timeout as silence since the latest accepted reduced member.
- Produces a partial final product when an incomplete stack times out.
- Renders previews only when enabled, membership has grown, and logical frame 1 is present so product names remain stable.

A failed preview does not advance the stored preview count, allowing a later poll to retry it.

## Finalization and retries

Finalization follows a deliberate order:

```text
claim attempt → write products → publish paths → mark terminal
```

The attempt and backoff are persisted before product work begins. A build or publish failure therefore leaves the stack active for a bounded retry, while exhausted attempts end in `error`.

Publishing happens before the terminal database update. A failure between those steps can produce a duplicate notification with identical paths, but cannot leave a stack marked complete without having notified the shipper.

## Process recovery

Ordinary exceptions are contained inside each worker loop. If a child dies violently—for example from an OOM kill—the supervisor exits and Docker restarts the service.

Because all lifecycle, retry, and preview state lives in PostgreSQL, workers restart without reconstructing process-local state. Claiming before product work also prevents a consistently failing stack from causing an endless restart loop.

## Scope

This PR adds the polling worker, crash-only supervisor, and site deployment configuration. PR7 adds cross-service E2E and recovery proof; PR8 adds structured lifecycle telemetry and final documentation.

Focused tests cover completion and timeout precedence, preview gating, finalization ordering, retries, publish failures, attempt exhaustion, worker-loop resilience, and supervisor behavior.

### DIFF
--- a/banzai/main.py
+++ b/banzai/main.py
@@ -26,7 +26,7 @@ from banzai.query import archive_get
 from banzai.utils import date_utils, stage_utils, import_utils, image_utils, fits_utils, file_utils
 from banzai.scheduling import (app, process_image, process_stackframe, requeue_missing_frames,
                                schedule_calibration_stacking)
-from banzai.stacking import validate_message
+from banzai import stacking
 from banzai.data import DataProduct
 from celery.schedules import crontab
 import celery
@@ -280,7 +280,7 @@ class StackframeListener(BanzaiQueueListener):
                          extra_tags={'error': str(e), 'body': repr(body)[:1000]})
             message.ack()
             return
-        if not validate_message(body):
+        if not stacking.validate_message(body):
             logger.error('Invalid message received, missing required fields',
                          extra_tags={'body': body})
             message.ack()
--- a/banzai/main.py
+++ b/banzai/main.py
@@ -305,6 +305,29 @@ def run_stackframe_worker():
     )
 
 
+def run_stacking_supervisor():
+    """Entry point for the smart-stacking supervisor."""
+    runtime_context = parse_args(
+        settings,
+        extra_console_arguments=[
+            {'args': ['--site-id'],
+             'kwargs': {'dest': 'site_id', 'required': True, 'help': 'Site identifier (e.g. lsc, ogg)'}},
+            {'args': ['--stack-retention-days'],
+             'kwargs': {'dest': 'stack_retention_days', 'type': int,
+                        'default': int(os.getenv('STACK_RETENTION_DAYS', 30)),
+                        'help': 'Days to retain terminal stacks before cleanup (default: 30)'}},
+            {'args': ['--stack-timeout-minutes'],
+             'kwargs': {'dest': 'stack_timeout_minutes', 'type': int,
+                        'default': int(os.getenv('STACK_TIMEOUT_MINUTES', 20)),
+                        'help': 'Minutes without a new stackframe before a partial stack is finalized (default: 20)'}},
+            {'args': ['--instrument-types'],
+             'kwargs': {'dest': 'instrument_types', 'default': os.getenv('INSTRUMENT_TYPES', '*'),
+                        'help': 'Comma-separated instrument types to stack, or * for all (default: *)'}},
+        ],
+    )
+    stacking.run_supervisor(runtime_context)
+
+
 def mark_frame(mark_as):
     parser = argparse.ArgumentParser(description="Set the is_bad flag to mark the frame as {mark_as}"
                                                  "for a calibration frame in the database ".format(mark_as=mark_as))
--- a/banzai/stacking.py
+++ b/banzai/stacking.py
@@ -1,9 +1,8 @@
-"""Smart stacking: per-camera polling worker + crash-only supervisor.
+"""Smartstack polling workers and their supervisor.
 
-The worker is stateless: every few seconds it polls the ``stacks`` table for one camera and
-does the obvious thing for each active stack (finalize when complete/timed out, otherwise
-render an incremental preview). All lifecycle state — backoff timers, attempt counts, preview
-progress — lives in the ``stacks`` row, never in process memory.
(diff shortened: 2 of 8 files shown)


banzai/main.py
```diff
@@ -26,7 +26,7 @@
 from banzai.utils import date_utils, stage_utils, import_utils, image_utils, fits_utils, file_utils
 from banzai.scheduling import (app, process_image, process_stackframe, requeue_missing_frames,
                                schedule_calibration_stacking)
-from banzai.stacking import validate_message
+from banzai import stacking
 from banzai.data import DataProduct
```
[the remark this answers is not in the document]
timbeccue: Moved CLI setup into main and passed the resulting context to the stacking supervisor. This removes the circular import without changing command behavior.

banzai/stacking.py
```diff
@@ -18,188 +26,201 @@
...
+    if stack_row.finalize_attempts >= MAX_FINALIZE_ATTEMPTS:
+        dbs.mark_stack_terminal(runtime_context.db_address, moluid, 'error')
+        logger.error('Smartstack exhausted finalize attempts; marking error',
+                     extra_tags={'moluid': moluid, 'finalize_attempts': stack_row.finalize_attempts})
+        return
+
+    dbs.claim_finalize_attempt(runtime_context.db_address, moluid, FINALIZE_BACKOFF_SECONDS)
+    try:
+        fits_path, small_thumbnail, large_thumbnail = smartstack_products.run_final(
+            stackframes, runtime_context, moluid)
```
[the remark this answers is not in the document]
timbeccue: The claim is intentionally outside the inner try. If it fails, no attempt or retry delay has been recorded, so the worker loop logs the error and tries again on its next poll. Once the claim succeeds, any later failure is handled here because the retry state is already saved.

banzai/stacking.py
```diff
@@ -18,188 +26,201 @@
...
+            {'args': ['--instrument-types'],
+             'kwargs': {'dest': 'instrument_types', 'default': os.getenv('INSTRUMENT_TYPES', '*'),
+                        'help': 'Comma-separated instrument types to stack, or * for all (default: *)'}},
+        ],
+    )
 
-    supervisor.start()
-    supervisor.monitor()
+    instrument_types = ([t.strip() for t in runtime_context.instrument_types.split(',')]
+                        if runtime_context.instrument_types != '*' else ['*'])
```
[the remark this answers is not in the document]
timbeccue: Normal Compose startup already converts empty/unset values to *. The rest is optional defensive coding; skipping for now to reduce complexity.

banzai/stacking.py
```diff
@@ -18,188 +25,180 @@
...
 
 
-def check_stack_complete(subframes, frmtotal):
+def check_stack_complete(stackframes, frmtotal):
     """Return True if the stack is ready to finalize.
 
-    Subframe rows are written only after reduction succeeds, so a stack is
-    complete when all expected rows are present or the instrument signalled
-    is_last.
+    Stackframe rows are written only after reduction succeeds, so a stack is complete when all
```
cmccully: Why is this an "or" here? Is this to catch cases where site software forgets to do is_last?

banzai/stacking.py
```diff
@@ -18,188 +25,180 @@
...
+    if instrument_types != ['*']:
+        instruments = [instrument for instrument in instruments if instrument.type in instrument_types]
+    cameras = [instrument.camera for instrument in instruments]
+    if not cameras:
+        logger.error('No cameras found at site; exiting so Docker restarts the container',
+                     extra_tags={'site_id': runtime_context.site_id})
+        sys.exit(1)
+
+    runtime_context_dict = vars(runtime_context)
+    processes = [multiprocessing.Process(target=run_worker_loop, args=(camera, runtime_context_dict), daemon=True)
```
cmccully, line 170: Is there any risk in having all the camera processes have the same parent process?

cmccully (approved):
Looks fine. Some of the comments read a little too much like AI.

banzai/stacking.py
timbeccue: Per our conversation, yes: one stack worker crashing results on all workers being restarted. This would result in an added delay if any other workers are processing preview or final data products. This design was chosen for simplicity and we'll leave it in the interest of having an MVP ready sooner to learn more about realistic failure modes.

banzai/stacking.py
timbeccue: Good point, this is probably unnecessarily defensive. Simplified to just rely on the last_frame flag and not check the total frame count.

timbeccue:
Cleaned up some of the code comments and addressed the other review suggestions.

PR #476, opened 2026-08-07
Smartstack: e2e tests, remove realtime containers

timbeccue:
## Summary

This is PR 7 of 8 completing the v1 banzai implementation of Smartstacking at site.

PR1 (#469): combine math
PR2 (#470): JPEG utilities
PR3 (#471): DB updates   
PR4 (#473): shipper contract                     
PR7 (#476): e2e tests, compose cleanup   &emsp;&emsp;&emsp;&emsp;    <-- you are here
PR8 (#477): logging

```text
PR1 combine math ───────┐
PR2 JPEG utilities ─────┤
PR4 shipper contract ───┘
```

PRs 1–4 provide the independent foundations for combining images, rendering JPEGs, storing stack state, and shipper integration (this PR). PR5 builds the data products, this PR adds the polling worker, crash-only supervisor, and site deployment configuration. PR7 adds cross-service E2E and recovery proof; PR8 adds structured lifecycle telemetry and final documentation.

## What's here

This branch has two changes: removing the realtime mode from the smartstack deployment (docker-compose-site.yml), and update the e2e tests to verify the behavior that was implemented works as intended.

### Remove realtime mode

`docker-compose-site.yml` becomes Smartstack-only:

- Removes the realtime `banzai-worker`, `banzai-large-worker`, and `banzai-listener` services.
- Removes their queue settings from the site defaults and E2E environment.
- Retains cache initialization, calibration download, stackframe ingress/reduction, and the stacking supervisor.

The realtime entry point and `docker-compose-local.yml` are unchanged. This changes only which pipeline the site deployment launches.

### End-to-end coverage
The site E2E suite now covers these concrete scenarios:

- **Single-frame smoke test:** Sends one raw stackframe through the site services and verifies that its reduced `e09` file is written and the stack reaches its expected terminal database state.

- **Cached calibration check:** Opens the reduced `e09` files and checks the calibration filenames recorded in their `L1IDBIAS`, `L1IDDARK`, and `L1IDFLAT` headers. Each referenced calibration must exist in the local cache, and its database record must point into that cache.

- **Normal three-frame stack:** Sends three raw stackframes and waits for previews after the first and second members. After the third member arrives, the test verifies:
  - one final `e45` FITS file and both JPEG sizes are written;
  - the previews and final product reuse the same first-frame-based paths;
  - `NCOMBINE=3` and `IMCOM001`–`IMCOM003` identify the three reduced inputs in order;
  - preview messages contain `fits: null` and the required thumbnail metadata; and
  - exactly one final shipper message contains the `e45` path.

- **Bad input does not block other stacks:** Corrupts one stack’s reduced `e09` file so final product generation repeatedly fails. That stack eventually reaches `error` after exhausting its retry attempts, while a healthy stack for the same camera still completes.

- **Worker restart recovery:** Force-kills the per-camera worker processes and verifies that the supervisor exits, Docker restarts the service, and a newly submitted stack completes successfully afterward.

The fixture uses isolated Smartstack queues and creates a dedicated shipper probe queue before sending any frames. This lets the tests inspect the actual RabbitMQ messages without interfering with another local deployment.

## Supporting changes

- `FINALIZE_BACKOFF_SECONDS` can be overridden for E2E while retaining the production default.
- Retention cleanup runs on the first worker tick and remains hourly afterward.
- The architecture guide is updated to describe the integrated PR1–PR6 behavior.

Focused combine, schema, product, publisher, and worker behavior remains owned by the earlier PRs; this PR proves those layers work together.


docs/smartstacking_architecture.md
```diff
@@ -1,358 +1,163 @@
...
+3. A Celery worker runs the normal BANZAI reduction and writes an `e09` file.
+4. After that succeeds, the worker saves the `e09` path and group information in PostgreSQL.
+5. A stacking process checks PostgreSQL about every five seconds. It opens the recorded `e09` files and makes a preview or final product when needed.
+6. BANZAI sends the product paths through RabbitMQ to the shipper.
 
-PostgreSQL is the source of truth. Redis notifications only mean:
+## Inputs and grouping
 
-> Something changed for this `MOLUID`; query the database again.
+The RabbitMQ ``banzai_stack_queue` message must contain `fits_file`, `last_frame`, and `instrument_enqueue_timestamp`. `last_frame` is stored in PostgreSQL as `is_last`.
```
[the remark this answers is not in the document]
timbeccue: fixed

docs/smartstacking_architecture.md
```diff
@@ -1,358 +1,163 @@
-# Smart Stacking Architecture
+# Smartstack Architecture Design
 
-**Implementation status:** orchestration scaffolding is in place. BANZAI reduces
-subframes, records them, detects complete groups, and marks those groups
-complete. It does **not yet create a stacked FITS file, JPEG preview, or archive
-upload**.
+Smartstacks are are exposures made from combining a series of shorter exposures. The benefits of this procedure is reduced tracking error, and the ability to provide previews of the stacking progress for realtime users as the stack is being processed. Since smartstacks require lots of files, creating the stacks in AWS risks saturating the bandwidth available to observatories. Therefore, smartstacks are designed to be processed with a banzai instance running direclty at site, and only the final smartstack products need to be sent to the archive.
```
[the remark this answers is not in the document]
timbeccue: fixed

banzai/tests/site_e2e/test_site_e2e.py
```diff
@@ -58,6 +60,70 @@ def check():
...
+def _corrupt_file(path):
+    """Overwrite an on-disk file that may be owned by a container UID with garbage.
+
+    The reduced e09 files are written from inside the site containers, so the host
+    test user may not own them. Overwrite through a throwaway container that mounts
+    the reduced-output tree at the same absolute path (mirrors
+    ``conftest._clean_data_dir``), which runs as root and always succeeds.
+    """
+    result = subprocess.run(
+        ["docker", "run", "--rm", "-v", f"{OUTPUT_DIR}:{OUTPUT_DIR}", "alpine",
```
cmccully, line 120: Spawning docker inside pytest still feels brittle to me. That's a pretty significant test dependency.

cmccully (approved):
I think things look mostly ok. I'm still not a fan having tests in pytest that have the "hidden" dependency of docker. I think I would still advocate for two testing approaches: either mock the responses of the external services or migrate the docker deploy testing to work with github actions to do the heavy lifting for the site deployment. 

I'm going to add Joey and Daniel as reviewers to decide if the docs are complete enough.

docs/smartstacking_architecture.md
```diff
@@ -1,358 +1,163 @@
-# Smart Stacking Architecture
+# Smartstack Architecture Design
 
-**Implementation status:** orchestration scaffolding is in place. BANZAI reduces
-subframes, records them, detects complete groups, and marks those groups
-complete. It does **not yet create a stacked FITS file, JPEG preview, or archive
-upload**.
+Smartstacks are exposures made from combining a series of shorter exposures. The benefits of this procedure are reduced tracking error and the ability to provide previews of the stacking progress for realtime users as the stack is being processed. Since smartstacks require lots of files, creating the stacks in AWS risks saturating the bandwidth available to observatories. Therefore, smartstacks are designed to be processed with a banzai instance running directly at site, and only the final smartstack products need to be sent to the archive.
```
jchate6: Should also mention the downsides like incompatibility with moving targets, loss of reproducability, possible (irreversible) ingestion of bad frames, increased readnoise (when applicable), etc. It's important for people to know when and when not to use this functionality based on their needs.

It should be pointed out that the previews will not be full data products but jpg snapshots.

I'd also change the last phrase to be "only the final smartstack products will be sent to the archive." The need for intermediate data products is not reduced by this process. We are choosing to neglect them in favor of saving bandwidth.

docs/smartstacking_architecture.md
```diff
@@ -1,358 +1,163 @@
...
+- A later successful reduction reopens a `complete` or `error` stack.
+- A `timeout` stack never reopens. Later reductions still write their `e09` files, but their database rows are rejected and they are not combined.
 
-Malformed JSON and messages missing required fields are logged, acknowledged,
-and discarded. Validation checks field presence only; FITS metadata is read
-later by the Celery task.
+After `STACK_RETENTION_DAYS`, cleanup removes the `stackframes` rows for old finished stacks. It also removes old `complete` and `error` stack rows. It keeps `timeout` stack rows so late frames continue to be rejected.
 
-### 2. Reduce and record
+Cleanup does not delete FITS or JPEG files.
```
jchate6: I'm a little confused by this. Cleanup doesn't remove e09's? just clears the DB?

docs/smartstacking_architecture.md
```diff
@@ -1,358 +1,163 @@
...
+For each build, BANZAI:
 
-Notification insertion uses an atomic Redis script. If the `MOLUID` is not
-already pending, the script increments `stack:notify:{camera}:sequence` and
-uses that sequence as the member's score. If the `MOLUID` is already pending,
-the script leaves its original score unchanged. Duplicate notifications are
-therefore coalesced without moving an older group behind newer work.
+1. Sorts the rows by frame number and opens the `e09` files.
+2. Uses the lowest-numbered file as the header and filename template.
+3. Combines matching `SCI` pixels in row sections to limit memory use, using a sum with 3-sigma rejection.
```
jchate6, line 118: Can we be be more specific with what "combines" means in this case? Is this a Median combine? An Add?

jchate6 (changes requested):
A few clarifications I'd like to see in the docs.

docs/smartstacking_architecture.md
```diff
@@ -1,358 +1,163 @@
...
 
-1. What problem does this subsystem solve?
-2. Which process owns each part of the work?
-3. Where is state stored, and what triggers a completion check?
-4. Which parts are still placeholders?
+The reduction level suffixes used:
 
-## The short version
+- `e00` is a raw exposure.
+- `e09` is one normally reduced stackframe exposure.
```
drhaz: What do you mean with "normally" ? Using the default BANZAI reduction step configuration, or a special limited one (e..g, w/o CR masking to ave time)?

docs/smartstacking_architecture.md
```diff
@@ -1,358 +1,163 @@
...
+    Reducer <-->|"read e00 and write e09"| Files
+    Stacker <-->|"read e09 and write products"| Files
+    Shipper -->|"open products"| Files
+```
 
-1. A listener receives a path to a raw subframe and sends reduction work to
-   Celery.
-2. After reduction succeeds, a per-camera stacking worker is notified. It
-   reloads the group from PostgreSQL and checks whether the stack is complete.
+1. Site software writes a raw `e00` FITS file and sends its absolute path to RabbitMQ `banzai_stack_queue`.
```
drhaz, line 52: The use of the e00 file, plus later statements on discarding postgress information what file was from a smart stack, and deferring clean up of files to other on -site processes:

What prevents someone finding that many e00 files were not shipped from the site to the archive and then manually requeuing those files to be sent to the archive.

docs/smartstacking_architecture.md
```diff
@@ -1,358 +1,163 @@
...
+2. The listener checks the message and publishes a Celery reduction task to Redis.
+3. A Celery worker runs the normal BANZAI reduction and writes an `e09` file.
```
drhaz: again, what is "normal"?

drhaz: Also, mention the use of locally cached master calibration files ... their management is not discussed here, is it?

drhaz (approved):
Documentation overall good, asking for a bit of clarification though. 

Beyond documentation, e00 fits file management is a concern raised in the comments.

docs/smartstacking_architecture.md
timbeccue: Thanks for these, I added most except for one I didn't understand. 

Can you explain more about the risk of ingesting bad frames? Does this just mean that we might send e45 stacks to the archive that include bad data and aren't useful, but will still live forever?

docs/smartstacking_architecture.md
timbeccue: Clarified that the files themselves are cleaned via a separate on-site process.

docs/smartstacking_architecture.md
timbeccue: End of the sentence says "using a sum with 3-sigma rejection"--is there a clearer way I should be wording that?

docs/smartstacking_architecture.md
timbeccue: Currently using the default ordered reduction stages (same that are used for normal exposure frames). I've clarified that in the docs. I suppose we could skip the CR masking since the stacking involves pixel rejection, I'll mention it to Curtis.

docs/smartstacking_architecture.md
timbeccue: Clarified, the "normal" usage, and added a section describing the calibration frame caching.

docs/smartstacking_architecture.md
timbeccue: I brought this up with Curtis and Mark and it sounds like we'll create a new obstype for subframes

> file name -n00.fits
> OBSTYPE SUB_EXP

docs/smartstacking_architecture.md
jchate6: I meant that if a frame is bad in some way (starlink interference, shutter issue, earthquake, etc) we won't keep it from being stacked along with everything else like a human would.

jchate6 (approved):
Thanks for the clarifications

PR #483, opened 2026-08-25
Use obstype SUB_EXP for raw and reduced stackframes

timbeccue:
Previously, smartstacking used raw stackframes with obstype=EXPOSE (e00), but made it more complicated to check whether raw frames that exist at site but not in the archive need to be requeued.

To separate concerns, we decided to change to using a new obstype SUB_EXP for smartstack stackframes. Raw stackframes are denoted with n00, and reduced frames use n09 (a simple change from the previous e00/e09). Final stacked data products and preview jpgs retain their e45/obstype=EXPOSE classification.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 Versions
 ========
 
+1.38.1 (2026-08-25)
+-------------------
+
+- Expect smartstack stackframes to have obstype=SUB_EXP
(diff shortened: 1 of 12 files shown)


timbeccue:
Closed by accident when deleting the temporary base branch. Identical PR reopened with #484.

PR #484, opened 2026-08-31
Use obstype SUB_EXP for raw and reduced stackframes

timbeccue:
Previously, smartstacking used raw stackframes with obstype=EXPOSE (e00), but made it more complicated to check whether raw frames that exist at site but not in the archive need to be requeued.

To separate concerns, we decided to change to using a new obstype SUB_EXP for smartstack stackframes. Raw stackframes are denoted with n00, and reduced frames use n09 (a simple change from the previous e00/e09). Final stacked data products and preview jpgs retain their e45/obstype=EXPOSE classification.

### DIFF
--- a/banzai/settings.py
+++ b/banzai/settings.py
@@ -49,8 +49,11 @@ CALIBRATION_SET_CRITERIA = {'BIAS': ['configuration_mode', 'binning'],
                             'READNOISE': ['configuration_mode', 'binning']}
 
 LAST_STAGE = {'BIAS': 'banzai.trim.Trimmer',
-              'DARK': 'banzai.uncertainty.PoissonInitializer', 'SKYFLAT': 'banzai.dark.DarkSubtractor',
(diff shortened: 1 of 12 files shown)


banzai/settings.py
```diff
@@ -50,7 +50,8 @@
 
 LAST_STAGE = {'BIAS': 'banzai.trim.Trimmer',
               'DARK': 'banzai.uncertainty.PoissonInitializer', 'SKYFLAT': 'banzai.dark.DarkSubtractor',
-              'SINISTRO': 'banzai.mosaic.MosaicCreator', 'STANDARD': None, 'EXPOSE': None, 'EXPERIMENTAL': None}
+              'SINISTRO': 'banzai.mosaic.MosaicCreator', 'STANDARD': None, 'EXPOSE': None, 'SUB_EXP': None,
```
cmccully: These should be obstype which is not SINISTRO.

cmccully (approved):
Overall looks good. Just a minor comment in the settings.py

banzai/settings.py
timbeccue: Done

